PR #10, opened 2025-01-20
Fix `requirements.txt` for MacOS, use Tensorflow 2.18.

hertschuh:
These changes are in line with what is done in Keras.


abheesht17 (approved):
Tried it locally, it works. Thanks!

PR #11, opened 2025-01-20
Add pre-commit hooks

abheesht17:
Currently, we have to manually run the formatting and linting scripts before pushing. Adding pre-commit hooks makes the scripts run automatically when you commit. Moreover, we can sync pre-commit with GitHub Actions.

This is a typical commit workflow:
```
(keras-rs-jax) abheesht-mac:keras-rs abheesht$ git add .
(keras-rs-jax) abheesht-mac:keras-rs abheesht$ git commit -m "Add pre-commit hooks"
format...................................................................Passed
lint.....................................................................Passed
[pre-commit-hooks 75c15e2] Add pre-commit hooks
 1 file changed, 12 insertions(+)
 create mode 100644 .pre-commit-config.yaml
(keras-rs-jax) abheesht-mac:keras-rs abheesht$ git remote -v
origin  https://github.com/abheesht17/keras-rs (fetch)
origin  https://github.com/abheesht17/keras-rs (push)
(keras-rs-jax) abheesht-mac:keras-rs abheesht$ git push
...
```

References: https://github.com/astral-sh/ruff-pre-commit?tab=readme-ov-file#using-ruff-with-pre-commit


.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
```
hertschuh, line 1: Do you know what it would take to make `api_gen` a pre-commit too?

It would make sense to have it first.

.github/workflows/actions.yml
```diff
@@ -70,8 +70,10 @@ jobs:
         run: |
           pip install -r requirements.txt --progress-bar off
           pip install --no-deps -e "." --progress-bar off
-      - name: Lint
-        run: bash shell/lint.sh
+      - name: Install pre-commit
+        run: pip install pre-commit
+      - name: Run pre-commit
+        run: pre-commit run --all-files
```
hertschuh, line 76: There is a difference between `format` and `lint` that got lost here.

`format` fixes the files. `lint` errors if anything needs to be changed.

In a PR, it's too late to fix the files, but we need to detect incorrectly formatted files and throw an error. Is it possible to do that?

`lint` runs each command and fails if any command fails. `format` runs all of them regardless and doesn't fail.

hertschuh (changes requested):

hertschuh:
> (keras-rs-jax) abheesht-mac:keras-rs abheesht$ git commit -m "Add pre-commit hooks"

What if the precommit hook fails? What happens? It doesn't commit?

What if you need to re-run it a few times as you're fixing the issues?

abheesht17:
> > (keras-rs-jax) abheesht-mac:keras-rs abheesht$ git commit -m "Add pre-commit hooks"
> 
> What if the precommit hook fails? What happens? It doesn't commit?
> 
> What if you need to re-run it a few times as you're fixing the issues?

Correct! Here is how it works:

```
git add .
git commit -m "Added awesome feature"
# Don't worry if errors are raised
# Your code is not compliant with ruff.
# ruff will automatically format the files so they might look different, but you'll need to stage the files
# again for committing.
# After fixing all errors:
git add .
git commit -m "Added feature"
```

.github/workflows/actions.yml
abheesht17: `format` in pre-commit will fail if the code is non-compliant and if it has to format the file.

abheesht17:
Will update the PR in a bit

.pre-commit-config.yaml
abheesht17: Did this too. The reason I did not do this initially is because it kept throwing an error. Had to look up some things - for example, pre-commit runs scripts concurrently, so had to ensure the `api_gen.sh` script runs serially (otherwise, copying from the repo to the temporary build directory errors out).

hertschuh:
> Will update the PR in a bit

Don't you need to add `pre-commit` to `requirements.txt` too (for local development)?

hertschuh (approved):

PR #12, opened 2025-01-22
Move project metadata from `setup.py` to `pyproject.toml`.

hertschuh:
Similar to what was done for Keras: https://github.com/keras-team/keras/pull/20427

Also:
- renamed `version_utils.py` to `version.py` for simplicity and consistency with Keras.
- removed `setup.cfg`, which was not used.
- made `pip_build.py` more robust.


abheesht17:
Could you please pull in the latest changes? Want to ensure `api_gen.sh` works fine after the pre-commit changes went in.

abheesht17 (approved):
LGTM!

PR #14, opened 2025-01-27
Add retrieval with data parallel training example

abheesht17:


examples/data_parallel_retrieval.py
```diff
@@ -0,0 +1,230 @@
...
+devices = jax.devices("gpu")  # Assume it has >1 local GPUs.
+data_parallel = keras.distribution.DataParallel(devices=devices)
+
+"""
+Alternatively, you can choose to create the `DataParallel` object
+using a 1D `DeviceMesh` object, like so:
+
+```
+mesh_1d = keras.distribution.DeviceMesh(
+    shape=(8,), axis_names=["data"], devices=devices
```
hertschuh, line 52: What is `8` here? Isn't it the number of devices? Can we put `len(devices)` instead?

hertschuh (commented):
Thanks for the example!

examples/data_parallel_retrieval.py
abheesht17: Oh, yeah. Missed this.

hertschuh (approved):

PR #15, opened 2025-01-27
Add `get_config` to `BruteForceRetrieval` layer.

hertschuh:
For consistency with other layers. Note that this does not serialize candidates.


abheesht17 (approved):
Looks good. Do you think we can remove the `name` arg (passed to `__init__`)? `kwargs` will subsume it

hertschuh:
> Looks good. Do you think we can remove the `name` arg (passed to `__init__`)? `kwargs` will subsume it

Oh, sure, I'll change it now

PR #20, opened 2025-01-30
Add GRU4Rec example

abheesht17:
Note: The loss in the TFRS example does not converge. So, changed a few things.


examples/sequential_retrieval.py
```diff
@@ -0,0 +1,422 @@
...
+neural network to model the sequential relationship. For more details,
+please refer to the [GRU4Rec](https://arxiv.org/abs/1511.06939) paper.
+
+Let's begin by importing all the necessary libraries, and setting the
+random seed for reproducibility.
+"""
+
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
```
hertschuh, line 17: Any reason to do that? At some point I was thinking we can write tests to run these automatically with each backend. But I never got around to doing that. But for it to work, the backend has to not be set within the file.

Did you test with TF and Torch too?

examples/sequential_retrieval.py
```diff
@@ -0,0 +1,422 @@
...
+import random
+
+import keras
+import pandas as pd
+import tensorflow as tf  # Needed only for the dataset
+from keras import ops
+
+import keras_rs
+
+keras.utils.set_random_seed(42)
```
hertschuh, line 29: Any reason to do that?

examples/sequential_retrieval.py
```diff
@@ -0,0 +1,422 @@
...
+LEARNING_RATE = 0.05
+
+"""
+## Dataset
+
+Next, we need to prepare our dataset. Like we did in the
+[basic retrieval](https://github.com/keras-team/keras-rs/blob/main/examples/basic_retrieval.py)
+example, we are going to use the MovieLens dataset. 
+
+The dataset preparation step is fairly involved. The original ratings dataset
```
hertschuh, line 69: Thanks for porting all the dataset massaging stuff here. That was most of the work I imagine.

examples/sequential_retrieval.py
```diff
@@ -0,0 +1,422 @@
...
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import collections
+import random
+
+import keras
+import pandas as pd
+import tensorflow as tf  # Needed only for the dataset
+from keras import ops
```
hertschuh, line 25: Further down, you're sometimes using `keras.ops.x` and sometimes `ops.x`. Make it consistent. I don't think this import is necessary.

hertschuh (commented):
Great work! There was a lot to do with the data processing.

examples/sequential_retrieval.py
abheesht17: I set the seed when I was testing training with different hyperparameters, etc. Can remove/keep it, doesn't matter which way

examples/sequential_retrieval.py
abheesht17: Yeah, pretty much. I was trying to not do the conversion to Pandas DF, and somehow use [`tf.data.group_by_window`](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#group_by_window). At some point, I gave up xD

examples/sequential_retrieval.py
abheesht17: Yep, working for all backends!

hertschuh (approved):

PR #21, opened 2025-02-06
Add DCN example

abheesht17:


examples/dcn.py
```diff
@@ -0,0 +1,522 @@
...
+| **Features / Label** | **Description**                | **Range** |
+|:--------------------:|:------------------------------:|:--------:|
+| $x_1$ = country      | Customer's resident country    | [0, 199] |
+| $x_2$ = bananas      | # bananas purchased            | [0, 23]  |
+| $x_3$ = cookbooks    | # cooking books purchased      | [0, 5]   |
+| y                    | Blender ad click likelihood    | -        |
+
+Then, we let the data follow the following underlying distribution:
+$y = f(x_1, x_2, x_3) = 0.1x_1 + 0.4x_2 + 0.7x_3 + 0.1x_1x_2 +$  
+$3.1x_2x_3 + 0.1x_3^2$.
```
hertschuh, line 166: I don't think this $ thing works on keras.io. Can you double check?

abheesht17: I saw it being used in other examples: https://github.com/keras-team/keras-io/blob/master/examples/nlp/multiple_choice_task_with_transfer_learning.py#L158

abheesht17: Oh, but it hasn't been rendered properly here: https://keras.io/examples/nlp/multiple_choice_task_with_transfer_learning/. Cool, will get rid of it.

abheesht17: https://github.com/keras-team/keras-io/blob/master/contributor_guide.md#mathematical-notation

hertschuh: Yeah, it's unfortunate. I'm not sure what this standard is, it's not markdown. Maybe there is a plugin to add support for it.

abheesht17: The solution right now is to add images.

abheesht17:
Results over 10 trials look good now:

```
DCN
RMSE = 0.7000066995620727 ± 0.07348885717921873; #params = 221953
===============
DCN with low-rank matrix
RMSE = 0.7314619958400727 ± 0.10956474137424495; #params = 192769
===============
Feedforward network
RMSE = 0.8033292055130005 ± 0.08073483901848993; #params = 221953
```

examples/dcn.py
abheesht17: Added it inside backticks. It won't be rendered as a LaTeX equation, but something's better than nothing

examples/dcn.py
```diff
@@ -0,0 +1,543 @@
...
+
+ratings_ds = ratings_ds.map(
+    lambda x, y: (
+        {
+            feature_name: lookup_layers[feature_name](x[feature_name])
+            for feature_name in vocabularies
+        },
+        y,
+    )
+)
```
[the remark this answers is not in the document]
abheesht17: Yeah, I was thinking of using `FeatureSpac`e, but thought this is pretty straightforward too, since we aren't doing any other preprocessing. Let me know if you think `FeatureSpace` will be better, I'll change it

examples/dcn.py
```diff
@@ -0,0 +1,543 @@
...
+    test_data,
+    model,
+    model_name=None,
+    num_trials=1,
+):
+    if model_name is not None:
+        print(model_name)
+
+    rmse_values = []
+    for _ in range(num_trials):
```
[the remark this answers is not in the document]
abheesht17: Oh, my bad. I moved `num_trials` inside the function and forgot to re-initialise the model for every trial.

Regarding its usefulness, it seems like it is. I am not getting DCN < opt DCN < deep network RMSE for every run. I'll run again though.

examples/dcn.py
```diff
@@ -0,0 +1,571 @@
...
+"""
+## Toy Example
+
+To illustrate the benefits of DCNs, let's consider a simple example. Suppose we
+have a dataset for modeling the likelihood of a customer clicking on a blender
+advertisement. The features and label are defined as follows:
+
+| **Features / Label** | **Description**                | **Range** |
+|:--------------------:|:------------------------------:|:--------:|
+| `x1` = country      | Customer's resident country    | [0, 199] |
```
hertschuh, line 176: Somehow it's still off between the horizontal line and the rest:
```
|:--------------------:|:------------------------------:|:--------:|
| `x1` = country      | Customer's resident country    | [0, 199] |
```

abheesht17: 🤔

examples/dcn.py
```diff
@@ -0,0 +1,571 @@
...
+        verbose=0,
+    )
+
+    results = model.evaluate(test_data, return_dict=True, verbose=0)
+    rmse_value = results["root_mean_squared_error"]
+
+    return rmse_value, model.count_params()
+
+
+def print_stats(rmse_list, num_params, model_name, num_trials=1):
```
hertschuh, line 153: Can't you just look at the length of `rmse_list` instead of having this extra `num_trials` argument.

The main reason is that 10 is hardcoded in 2 places and it's not intuitive why you're passing 10 to `print_stats` further down.

hertschuh (approved):

PR #22, opened 2025-02-11
Add `DotInteraction` layer

abheesht17:


keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -0,0 +1,131 @@
...
+        skip_gather: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+
+        # Passed args.
+        self.self_interaction = self_interaction
+        self.skip_gather = skip_gather
+
+    def call(self, inputs: List[types.Tensor]) -> types.Tensor:
```
abheesht17, line 47: Are we okay with `List[types.Tensor]`?

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -0,0 +1,131 @@
...
+            `[batch_size, k]` where `k` is
+            `num_features * num_features` if `skip_gather` is `True`. Otherwise,
+            `k` is `num_features * (num_features + 1) / 2` if
+            `self_interaction` is `True`, and
+            `num_features * (num_features - 1) / 2` if not.
+        """
+
+        # A `ValueError` will be raised if the feature tensors inside the list
+        # are not of equal shape. So, we do not need to handle the error here.
+        features = ops.array(inputs)
```
abheesht17, line 66: This or `ops.concatenate([ops.expand_dims(x, axis=1) for x in inputs], axis=1)`?

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+        else:
+            activations = pairwise_interaction_matrix[tril_mask]
+
+        # Rank-2 tensor.
+        activations = ops.reshape(activations, (batch_size, -1))
+
+        return activations
+
+    def compute_output_shape(
```
abheesht17, line 104: Do you think we will not need to implement this if we explicitly mention the shape on line 100 instead of `-1`?

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -0,0 +1,131 @@
+from typing import Any, List, Tuple
```
hertschuh, line 1: Please use `list` instead of `List` and `tuple` instead of `Tuple` for type annotations for consistency. So you only need to import `Any`.

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(
+        self,
+        self_interaction: bool = False,
+        skip_gather: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+
+        # Passed args.
```
hertschuh, line 43: nitpick: I would say "# Attributes." instead of "# Passed args."

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -0,0 +1,131 @@
...
+            activations = pairwise_interaction_matrix[tril_mask]
+
+        # Rank-2 tensor.
+        activations = ops.reshape(activations, (batch_size, -1))
+
+        return activations
+
+    def compute_output_shape(
+        self, input_shape: List[Tuple[int, int]]
+    ) -> Tuple[int, int]:
```
hertschuh, line 106: Use `types.TensorShape` instead of `Tuple[int, int]`. Part of it is that the dimensions can be `None`, in particular the batch size. Although we would lose the information that the shape is always 2D.

keras_rs/src/layers/modeling/dot_interaction_test.py
```diff
@@ -0,0 +1,101 @@
...
+        self.f33 = ops.dot(feature3[0], feature3[0])
+
+    def test_layer_call_1(self):
+        layer = DotInteraction(self_interaction=False, skip_gather=False)
+        output = layer(self.input)
+        print(f"{output=}")
+        print(ops.array([[self.f12, self.f13, self.f23]]))
+        self.assertAllClose(output, ops.array([[self.f12, self.f13, self.f23]]))
+
+        # def test_layer_call_2(self):
```
hertschuh, line 37: Uncomment?

keras_rs/src/layers/modeling/dot_interaction_test.py
```diff
@@ -0,0 +1,101 @@
...
+            feature1 = keras.layers.Input(shape=(5,))
+            feature2 = keras.layers.Input(shape=(5,))
+            feature3 = keras.layers.Input(shape=(5,))
+            x = DotInteraction()([feature1, feature2, feature3])
+            x = keras.layers.Dense(units=1)(x)
+            model = keras.Model([feature1, feature2, feature3], x)
+            return model
+
+        self.run_model_saving_test(
+            model=get_model(),
```
hertschuh, line 99: You can just remove line 89 and line 96 and just pass `model`.

keras_rs/src/layers/modeling/dot_interaction_test.py
```diff
@@ -0,0 +1,101 @@
...
+        feature3 = self.input[2]
+
+        self.f11 = ops.dot(feature1[0], feature1[0])
+        self.f12 = ops.dot(feature1[0], feature2[0])
+        self.f13 = ops.dot(feature1[0], feature3[0])
+        self.f22 = ops.dot(feature2[0], feature2[0])
+        self.f23 = ops.dot(feature2[0], feature3[0])
+        self.f33 = ops.dot(feature3[0], feature3[0])
+
+    def test_layer_call_1(self):
```
hertschuh, line 30: Can you rename the tests like `test_self_interaction_false_skip_gather_false` etc.

hertschuh (changes requested):
Thanks for the PR!

keras_rs/src/layers/modeling/dot_interaction.py
hertschuh: Yes, as `list[types.Tensor]`

hertschuh (approved):

PR #25, opened 2025-02-12
Add ClippyAdagrad optimizer.

hertschuh:
An Adagrad variant with adaptive clipping.


hertschuh:
> @hertschuh - are we moving this to Keras? Should I wait before reviewing?

Do not review now. I'm looking into augmenting the Keras Adagrad optimizer instead.

PR #28, opened 2025-02-14
Fix dot interaction

abheesht17:
- Using `tril_mask` to pick the lower triangular part of the matrix does not work for JAX when `skip_gather` is `False`. So, we generate indices and then index into the array.
- Cannot compare shapes directly.

### DIFF
--- a/keras_rs/src/layers/modeling/dot_interaction.py
+++ b/keras_rs/src/layers/modeling/dot_interaction.py
@@ -45,6 +45,44 @@ class DotInteraction(keras.layers.Layer):
         self.self_interaction = self_interaction
         self.skip_gather = skip_gather
 
+    def _generate_tril_mask(
+        self, pairwise_interaction_matrix: types.Tensor
+    ) -> types.Tensor:
+        """Generates lower triangular mask."""
+
+        # If `self.self_interaction` is `True`, keep the main diagonal.
+        k = -1
+        if self.self_interaction:
+            k = 0
+
+        # Typecast k from Python int to tensor, because `ops.tril` uses
+        # `tf.cond` (which requires tensors).
+        # TODO (abheesht): Remove typecast once fix is merged in core Keras.
+        if keras.config.backend() == "tensorflow":
+            k = ops.array(k)
+        tril_mask = ops.tril(
+            ops.ones_like(pairwise_interaction_matrix, dtype=bool),
+            k=k,
+        )
+
+        return tril_mask
+
+    def _get_lower_triangular_indices(self, num_features: int) -> list[int]:
+        """Python function which generates indices to get the lower triangular
+        matrix as if it were flattened.
+        """
+        flattened_indices = []
+        for i in range(num_features):
+            k = i
+            # if `self.self_interaction` is `True`, keep the main diagonal.
+            if self.self_interaction:
+                k += 1
+            for j in range(k):
+                flattened_index = i * num_features + j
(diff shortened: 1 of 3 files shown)


keras_rs/src/utils/keras_utils.py
```diff
@@ -25,3 +27,16 @@ def clone_initializer(
         return initializer_class.from_config(config)
     # If we get a string or dict, just return as we cannot and should not clone.
     return initializer
+
+
+def check_shape(shape1: types.TensorShape, shape2: types.TensorShape) -> bool:
```
hertschuh, line 32: Let's give it a slightly more descriptive name, maybe `check_shapes_compatible`? Or `static_dimensions_equal`?

Also add a brief docstring.

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -64,52 +103,50 @@ def call(self, inputs: list[types.Tensor]) -> types.Tensor:
...
         # Else, "pick" only the lower triangle entries.
         if self.skip_gather:
+            tril_mask = self._generate_tril_mask(pairwise_interaction_matrix)
+
             activations = ops.multiply(
                 pairwise_interaction_matrix,
                 ops.cast(tril_mask, dtype=pairwise_interaction_matrix.dtype),
             )
+            # Rank-2 tensor.
+            activations = ops.reshape(activations, (batch_size, -1))
```
hertschuh, line 142: I've been avoiding using `-1` in reshape when it is possible to calculate the explicit value. The reason is that on Tensorflow, when you have a dynamic batch size (which is often the case with a tf dataset), `(batch_size, -1)` will make the second dimension dynamic too even though it can be computed statically. Then, the output will be fully dynamic with a static shape of `(None, None)`.

Instead, put the explicit value (I believe it's simply `num_features * num_features`). Note that this also works if `num_features` happens to be dynamic.

If we want to verify that what I'm saying is actually true :) here is how you test it:
https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/stft_spectrogram_test.py#L294

keras_rs/src/layers/modeling/dot_interaction.py
```diff
@@ -64,52 +103,50 @@ def call(self, inputs: list[types.Tensor]) -> types.Tensor:
...
         else:
-            activations = pairwise_interaction_matrix[tril_mask]
-
-        # Rank-2 tensor.
-        activations = ops.reshape(activations, (batch_size, -1))
+            flattened_indices = self._get_lower_triangular_indices(num_features)
+            activations = ops.take(
+                ops.reshape(pairwise_interaction_matrix, (batch_size, -1)),
```
hertschuh, line 146: Same comment about `-1` -> `num_features * num_features`.

hertschuh (changes requested):
Thanks for the fix!

keras_rs/src/layers/modeling/dot_interaction.py
abheesht17: Interesting. `None * None` will work?

hertschuh: Oh, what I meant is that if it's dynamic, `ops.shape` will return a tensor of rank 0 for that dimension, multiplying tensors like that is allowed, and then `ops.reshape` accepts dimensions that are dynamic and specified via a tensor.

hertschuh (approved):

PR #29, opened 2025-02-17
Add multi-task (retrieval + ranking) example

abheesht17:
Delta from TFRS example:

- Data processing, model definition as in the basic ranking/retrieval tutorials (except for a few changes).
- Prediction at the bottom: TFRS shows a rating prediction only. In this example, we have done retrieval first, and then done a rating on the retrieved set.

### DIFF
--- a/examples/multi_task.py
+++ b/examples/multi_task.py
@@ -1,5 +1,5 @@
 """
-# Multi-task Recommenders
+# Multi-task Recommenders: Retrieval + Ranking
 
 In the
 [basic retrieval](https://github.com/keras-team/keras-rs/blob/main/examples/basic_retrieval.py)
--- a/examples/multi_task.py
+++ b/examples/multi_task.py
@@ -1,0 +1,389 @@
+"""
+# Multi-task Recommenders
+
+In the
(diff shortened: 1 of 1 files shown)


examples/multi_task.py
```diff
@@ -0,0 +1,389 @@
...
+)
+model.compile(optimizer=tf.keras.optimizers.Adagrad(0.1))
+model.fit(train_ratings, epochs=5)
+
+model.evaluate(test_ratings)
+
+"""
+Let's plot a table of the metrics and pen down our observations:
+
+| Model                 | Top-K Accuracy (↑) | RMSE (↓) |
```
hertschuh, line 340: Do the little up / down arrows in the table header do anything?

hertschuh (commented):
LGTM. Thanks!

Just one question out of curiosity.

examples/multi_task.py
abheesht17: It just means that a higher value of accuracy is better, and a lower value of RMSE is better

hertschuh: Oh ok

hertschuh (approved):

PR #30, opened 2025-02-17
Add ScANN example

abheesht17:
Omitted

- Accuracy computation
- Tuning ScANN

Should we include these?

My question is: if we want to include these, we might have to manually iterate over the test dataset and compute predictions. Might be a little tedious. Let me know if we want to do this, @hertschuh.

### DIFF
--- a/examples/scann.py
+++ b/examples/scann.py
@@ -1,5 +1,5 @@
 """
-# Faster retrieval with ScANN
+# Faster retrieval with Scalable Nearest Neighbours (ScANN)
 
 Retrieval models are designed to quickly identify a small set of highly relevant
 candidates from vast pools of data, often comprising millions or even hundreds
--- a/examples/scann.py
+++ b/examples/scann.py
@@ -1,0 +1,261 @@
+"""
+# Faster retrieval with ScANN
+
+Retrieval models are designed to quickly identify a small set of highly relevant
+candidates from vast pools of data, often comprising millions or even hundreds
+of millions of items. To effectively respond to the user's context and behavior
+in real time, these models must perform this task in just milliseconds.
+
+Approximate nearest neighbor (ANN) search is the key technology that enables
+this level of efficiency. In this tutorial, we'll demonstrate how to leverage
+ScANN—a cutting-edge nearest neighbor retrieval library—to effortlessly scale
+retrieval for millions of items.
+
+[ScANN](https://research.google/blog/announcing-scann-efficient-vector-similarity-search/),
+developed by Google Research, is a high-performance library designed for
+dense vector similarity search at scale. It efficiently indexes a database of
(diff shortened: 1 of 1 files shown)


examples/scann.py
```diff
@@ -0,0 +1,261 @@
...
+Note: ScANN does not have its own separate layer in KerasRS because the ScANN
+library is TensorFlow-only. Here, in this example, we directly use the ScANN
+library and demonstrate its usage with KerasRS.
+
+## Imports
+
+Let's install the `scann` library and import all necessary packages.
+"""
+
+# ruff: noqa: E402
```
hertschuh, line 37: Do you need this? The imports look like they're at the top.

examples/scann.py
```diff
@@ -0,0 +1,261 @@
...
+[ScANN](https://research.google/blog/announcing-scann-efficient-vector-similarity-search/),
+developed by Google Research, is a high-performance library designed for
+dense vector similarity search at scale. It efficiently indexes a database of
+candidate embeddings, enabling rapid search during inference. By leveraging
+advanced vector compression techniques and finely tuned algorithms, ScaNN
+strikes an optimal balance between speed and accuracy. As a result, it can
+significantly outperform brute-force search methods, delivering fast retrieval
+with minimal loss in accuracy.
+
+We will start with copying over code from the
```
hertschuh, line 23: Nitpick: instead of saying "We will start with copying over code from the", say "We will start with the same code as the".

The former makes it sound like the code is running a `cp` command or something.

examples/scann.py
```diff
@@ -0,0 +1,261 @@
...
+        candidate_embeddings
+        * keras.random.uniform(keras.ops.shape(candidate_embeddings))
+        for _ in range(100)
+    ],
+    axis=0,
+)
+
+user_embedding = model.user_embedding(keras.ops.array([10, 5, 42, 345]))
+
+print(f"{candidate_embeddings.shape=}", f"{user_embedding.shape=}")
```
hertschuh, line 182: Is it intentional that this `print` is here?

hertschuh:
> Omitted
> 
> * Accuracy computation
> * Tuning ScANN
> 
> Should we include these?
> 
> My question is: if we want to include these, we might have to manually iterate over the test dataset and compute predictions. Might be a little tedious. Let me know if we want to do this, @hertschuh.

I don't think we need these.

examples/scann.py
abheesht17: For some reason, the `"""shell` block isn't considered a string by ruff, and it keeps throwing E402.

examples/scann.py
abheesht17: Oh, sorry. I was using it for debugging

hertschuh (approved):
Thanks!

PR #32, opened 2025-02-24
Add pairwise losses (MSE, Hinge, Logistic, SoftZeroOne)

abheesht17:
To be discussed offline:
- Currently, we are passing the mask explicitly with `y_true`. Can use `sample_weight`, but have a few questions about that [keep it for now].
- Do we need to do the up-ranking for rank-1 tensors? Or can we ignore? [keep `dim=0` expansion, but check TFR]
- Doc-strings: should we remove args such as reduction, name? [keep them]
- Should we set doc-strings by setting `cls.__doc__` to avoid repetition? [yes]

TODO:

- Add `ops.stop_gradient()`.
- Add temperature, etc. as args (can be done in another PR).


keras_rs/src/utils/pairwise_loss_utils.py
```diff
@@ -0,0 +1,100 @@
...
+        ops.expand_dims(x, axis=-1),
+        ops.expand_dims(x, axis=-2),
+    )
+
+
+def pairwise_comparison(
+    labels: types.Tensor,
+    logits: types.Tensor,
+    mask: types.Tensor,
+    logits_op: ops,
```
hertschuh, line 20: The type annotation to `logits_op` seems wrong. Shouldn't it be a `Callable[[Tensor, Tensor], Tensor]`?

keras_rs/src/utils/pairwise_loss_utils.py
```diff
@@ -0,0 +1,100 @@
...
+        y_true_shape, mask_shape
+    ):
+        raise ValueError(
+            "`y_true['labels']` and `y_true['mask']` should have the same "
+            f"shape. Received: `y_true['labels'].shape` = {y_true_shape}, "
+            f"`y_true['mask'].shape` = {mask_shape}."
+        )
+
+    if y_true_rank == 1:
+        y_true = ops.expand_dims(y_true, axis=0)
```
hertschuh, line 95: I'm surprised the extra dimension as added at index 0 rather than at the end (-1). How come?

keras_rs/src/utils/pairwise_loss_utils.py
```diff
@@ -0,0 +1,100 @@
...
+) -> tuple[types.Tensor, types.Tensor, Optional[types.Tensor]]:
+    """
+    This utility function does three things:
+
+    - Checks that `y_true`, `y_pred` are of rank 1 or 2.
+    - Check that `y_true`, `y_pred`, `mask` have the same shape.
+    - Add batch dimension if rank = 1.
+    """
+
+    def get_shape(x: types.Tensor) -> tuple[types.TensorShape, int]:
```
hertschuh, line 54: I don't think this is worth the trouble (and the name is a bit confusing).

You can just do `len(y_true_shape)` where needed.

keras_rs/src/losses/pairwise_loss.py
```diff
@@ -0,0 +1,93 @@
...
+    correctly order items within each list. Any pairwise loss function computes
+    the loss value by comparing pairs of items within each list, penalizing
+    cases where an item with a higher true label has a lower predicted score
+    than an item with a lower true label.
+
+    In order to implement any kind of pairwise loss, override the
+    `pairwise_loss` method.
+    """
+
+    # TODO: Add `temperature`, `lambda_weights`.
```
hertschuh, line 25: Was this in the original implementation? Can you create a github issue for this?

keras_rs/src/losses/pairwise_loss.py
```diff
@@ -0,0 +1,93 @@
...
+
+import keras
+from keras import ops
+
+from keras_rs.src import types
+from keras_rs.src.utils.pairwise_loss_utils import pairwise_comparison
+from keras_rs.src.utils.pairwise_loss_utils import process_loss_call_inputs
+
+
+class PairwiseLoss(keras.losses.Loss):
```
hertschuh, line 12:
```python
class PairwiseLoss(keras.losses.Loss, abc.ABC):
```

hertschuh (commented):
First pass at the super class and utils.

keras_rs/src/utils/pairwise_loss_utils.py
abheesht17: Yep, this is one thing I wanted to discuss! Let's discuss offline

keras_rs/src/losses/pairwise_loss.py
abheesht17: Yep, it's there in the original implementation

keras_rs/src/utils/pairwise_loss_utils.py
abheesht17: Yep, need to change this

keras_rs/src/losses/pairwise_loss.py
```diff
@@ -0,0 +1,136 @@
...
+        self,
+        y_true: types.Tensor,
+        y_pred: types.Tensor,
+    ) -> types.Tensor:
+        """
+        Args:
+            y_true: tensor or dict. Ground truth values. If tensor, of shape
+                `(list_size)` for unbatched inputs or `(batch_size, list_size)`
+                for batched inputs. If an item has a label of -1, it is ignored
+                in loss computation.If it is a dictionary, it should have two
```
hertschuh, line 66: nitpick: space between computation and If

keras_rs/src/losses/pairwise_hinge_loss.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@keras_rs_export("keras_rs.losses.PairwiseHingeLoss")
+class PairwiseHingeLoss(PairwiseLoss):
+    def pairwise_loss(self, pairwise_logits: types.Tensor) -> types.Tensor:
+        return ops.relu(ops.subtract(ops.array(1), pairwise_logits))
+
+
+PairwiseHingeLoss.__doc__ = (
+    pairwise_loss_subclass_doc_string.replace(
```
hertschuh, line 16: I've seen the `replace` approach used in Keras, but I think the `format` approach is a bit cleaner: https://github.com/keras-team/keras/blob/master/keras/src/applications/mobilenet_v2.py#L499

keras_rs/src/losses/pairwise_hinge_loss.py
```diff
@@ -0,0 +1,34 @@
...
+        "{{formula}}",
+        """
+        loss = sum_{i} sum_{j} I(y_i > y_j) * max(0, 1 - (s_i - s_j))
+        """,
+    )
+    .replace(
+        "{{explanation}}",
+        """\033[A
```
hertschuh, line 24: Does that actually work? Is it just because the dash `-` has to come right after the `"""`?

keras_rs/src/losses/pairwise_loss.py
```diff
@@ -0,0 +1,136 @@
...
+        losses, weights = self.compute_unreduced_loss(
+            labels=y_true, logits=y_pred, mask=mask
+        )
+        losses = ops.multiply(losses, weights)
+        losses = ops.sum(losses, axis=-1)
+        return losses
+
+
+pairwise_loss_subclass_doc_string = (
+    "    Computes pairwise hinge loss between true labels and predicted scores."
```
hertschuh, line 97: Shouldn't this simply be in the triple quotes?

keras_rs/src/losses/pairwise_hinge_loss.py
```diff
@@ -0,0 +1,34 @@
...
+      - `max(0, 1 - (s_i - s_j))` is the hinge loss, which penalizes cases where
+        the score difference `s_i - s_j` is not sufficiently large when
+        `y_i > y_j`.
+        """,
+    )
+    .replace(
+        "{{extra_args}}",
+        "\033[A",
```
hertschuh, line 32: Does that work for removing the empty line?

keras_rs/src/losses/pairwise_hinge_loss_test.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_model_fit(self):
+        inputs = keras.Input(shape=(20,), dtype="float32")
+        outputs = keras.layers.Dense(5)(inputs)
+        model = keras.Model(inputs=inputs, outputs=outputs)
+
+        model.compile(loss=PairwiseHingeLoss(), optimizer="adam")
+        model.fit(
+            x=keras.random.normal((2, 20)),
+            y=keras.random.randint((2, 5), minval=0, maxval=2),
+        )
```
hertschuh, line 107: Add a test for serialization and/or saving?

hertschuh (commented):
Thanks for the updates!

The comments for pairwise hinge loss apply to the other ones too.

keras_rs/src/losses/pairwise_hinge_loss.py
abheesht17: Yep, it works!

```
Computes pairwise hinge loss between true labels and predicted scores.

    This loss function is designed for ranking tasks, where the goal is to
    correctly order items within each list. It computes the loss by comparing
    pairs of items within each list, penalizing cases where an item with a
    higher true label has a lower predicted score than an item with a lower
    true label.

    For each list of predicted scores `s` in `y_pred` and the corresponding list
    of true labels `y` in `y_true`, the loss is computed as follows:

    ```
    loss = sum_{i} sum_{j} I(y_i > y_j) * (s_i - s_j)^2
    ```

    where:
      - `y_i` and `y_j` are the true labels of items `i` and `j`, respectively.
      - `s_i` and `s_j` are the predicted scores of items `i` and `j`,
        respectively.
      - `I(y_i > y_j)` is an indicator function that equals 1 if `y_i > y_j`,
        and 0 otherwise.
      - `(s_i - s_j)^2` is the squared difference between the predicted scores
        of items `i` and `j`, which penalizes discrepancies between the
        predicted order of items relative to their true order.
        
    Args:
        reduction: Type of reduction to apply to the loss. In almost all cases
            this should be `"sum_over_batch_size"`. Supported options are
            `"sum"`, `"sum_over_batch_size"`, `"mean"`,
            `"mean_with_sample_weight"` or `None`. `"sum"` sums the loss,
            `"sum_over_batch_size"` and `"mean"` sum the loss and divide by the
            sample size, and `"mean_with_sample_weight"` sums the loss and
            divides by the sum of the sample weights. `"none"` and `None`
            perform no aggregation. Defaults to `"sum_over_batch_size"`.
        name: Optional name for the loss instance.
        dtype: The dtype of the loss's computations. Defaults to `None`, which
            means using `keras.backend.floatx()`. `keras.backend.floatx()` is a
            `"float32"` unless set to different value
            (via `keras.backend.set_floatx()`). If a `keras.DTypePolicy` is
            provided, then the `compute_dtype` will be utilized.
```

hertschuh: What is the thing you pasted here?

Does it work with IDEs (e.g. VSCode for code completion and when you mouse over a symbol)

Does it work on Keras.io (the thing that generates the documentation)?

abheesht17: > What is the thing you pasted here?

```
print(PairwiseMeanSquaredError.__doc__)
```

> Those special codes are a terminal concept. So it works when you print it. But outside of a terminal, it's unlikely to work, there is no equivalent concept in markdown, html...

Ah, I see

keras_rs/src/losses/pairwise_loss.py
```diff
@@ -0,0 +1,137 @@
...
+    where:
+      - `y_i` and `y_j` are the true labels of items `i` and `j`, respectively.
+      - `s_i` and `s_j` are the predicted scores of items `i` and `j`,
+        respectively.
+      - `I(y_i > y_j)` is an indicator function that equals 1 if `y_i > y_j`,
+        and 0 otherwise.
+      {explanation}
+
+    Args:
+        {extra_args}
```
hertschuh, line 121: It looks like `{extra_args}` is never actually used (always empty).

keras_rs/src/losses/pairwise_mean_squared_error.py
```diff
@@ -0,0 +1,71 @@
...
+        # Include only valid pairs.
+        pairwise_weights = ops.multiply(
+            pairwise_weights, ops.cast(valid_pair, dtype=pairwise_weights.dtype)
+        )
+
+        return pairwise_mse, pairwise_weights
+
+
+formula = "loss = sum_{i} sum_{j} I(y_i > y_j) * (s_i - s_j)^2"
+explanation = """\033[A
```
hertschuh, line 61: Remove the two `\033[A` and just use `"` instead `"""`.

keras_rs/src/losses/pairwise_soft_zero_one_loss.py
```diff
@@ -0,0 +1,31 @@
...
+    def pairwise_loss(self, pairwise_logits: types.Tensor) -> types.Tensor:
+        return ops.where(
+            ops.greater(pairwise_logits, ops.array(0.0)),
+            ops.subtract(ops.array(1.0), ops.sigmoid(pairwise_logits)),
+            ops.sigmoid(ops.negative(pairwise_logits)),
+        )
+
+
+formula = "loss = sum_{i} sum_{j} I(y_i > y_j) * (1 - sigmoid(s_i - s_j))"
+explanation = """\033[A
```
hertschuh, line 20: Remove the two `\033[A` and just use `"` instead `"""`.

keras_rs/src/losses/pairwise_hinge_loss.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@keras_rs_export("keras_rs.losses.PairwiseHingeLoss")
+class PairwiseHingeLoss(PairwiseLoss):
+    def pairwise_loss(self, pairwise_logits: types.Tensor) -> types.Tensor:
+        return ops.relu(ops.subtract(ops.array(1), pairwise_logits))
+
+
+formula = "loss = sum_{i} sum_{j} I(y_i > y_j) * max(0, 1 - (s_i - s_j))"
+explanation = """\033[A
```
hertschuh, line 16: Remove the two `\033[A` and just use `"` instead `"""`.

keras_rs/src/losses/pairwise_logistic_loss.py
```diff
@@ -0,0 +1,35 @@
...
+                ops.add(
+                    ops.array(1),
+                    ops.exp(ops.negative(ops.abs(pairwise_logits))),
+                )
+            ),
+        )
+
+
+formula = "loss = sum_{i} sum_{j} I(y_i > y_j) * log(1 + exp(-(s_i - s_j)))"
+explanation = """\033[A
```
hertschuh, line 24: Remove the two `\033[A` and just use `"` instead `"""`.

keras_rs/src/losses/pairwise_loss.py
abheesht17: Yeah, keeping it in case we need to add other losses which might have extra args

keras_rs/src/losses/pairwise_mean_squared_error.py
abheesht17: Did it slightly differently, please take a look

hertschuh (approved):

PR #35, opened 2025-02-28
Add Sampling Probability Correction layer.

hertschuh:
This layer corrects logits based on sampling probability.

### DIFF
--- a/keras_rs/src/layers/retrieval/sampling_probability_correction.py
+++ b/keras_rs/src/layers/retrieval/sampling_probability_correction.py
@@ -1,0 +1,48 @@
+from typing import Any
+
+import keras
+from keras import ops
+
+from keras_rs.src import types
+from keras_rs.src.api_export import keras_rs_export
+
+
+@keras_rs_export("keras_rs.layers.SamplingProbabilityCorrection")
+class SamplingProbabilityCorrection(keras.layers.Layer):
+    """Sampling probability correction.
+
+    Corrects the logits to reflect the sampling probability of negatives.
+
+    Args:
+        epsilon: float. Small float added to sampling probability to avoid
+            taking the log of zero. Defaults to 1e-6.
+        **kwargs: Args to pass to the base class.
+    """
+
+    def __init__(self, epsilon: float = 1e-6, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.epsilon = epsilon
+        self.built = True
+
+    def call(
+        self,
+        logits: types.Tensor,
+        candidate_sampling_probability: types.Tensor,
+    ) -> types.Tensor:
+        """Corrects input logits to account for candidate sampling probability.
+
+        Args:
+            logits: The logits to correct.
+            candidate_sampling_probability: The sampling probability.
+
+        Returns:
+            The corrected logits.
+        """
+        return logits - ops.log(
+            ops.clip(candidate_sampling_probability, self.epsilon, 1.0)
+        )
+
+    def get_config(self) -> dict[str, Any]:
+        config: dict[str, Any] = super().get_config()
+        config.update({"epsilon": self.epsilon})
+        return config
--- a/keras_rs/src/layers/retrieval/sampling_probability_correction_test.py
+++ b/keras_rs/src/layers/retrieval/sampling_probability_correction_test.py
@@ -1,0 +1,68 @@
+import keras
+from keras import ops
+from keras.layers import deserialize
+from keras.layers import serialize
+
+from keras_rs.src import testing
+from keras_rs.src.layers.retrieval import sampling_probability_correction
+
+
+class SamplingProbabilityCorrectionTest(testing.TestCase):
+    def setUp(self):
+        shape = (10, 20)  # (num_queries, num_candidates)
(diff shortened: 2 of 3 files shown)


keras_rs/src/layers/retrieval/sampling_probability_correction.py
```diff
@@ -0,0 +1,41 @@
...
+from keras_rs.src import types
+from keras_rs.src.api_export import keras_rs_export
+
+
+@keras_rs_export("keras_rs.layers.SamplingProbablityCorrection")
+class SamplingProbablityCorrection(keras.layers.Layer):
+    """Sampling probability correction.
+
+    Args:
+        epsilon: Small float added to sampling probability to avoid taking the
```
abheesht17, line 15: I've been thinking about our doc-strings. In the rest of Keras, the format is `{arg_name}: {dtype}. {description}`. For example, https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/f_net/f_net_backbone.py#L41.
I wonder if we should do the same. Or because we are doing static typing, it'll show up in the generated documentation anyway?

keras_rs/src/layers/retrieval/sampling_probability_correction.py
```diff
@@ -0,0 +1,41 @@
...
+            log of zero. Defaults to 1e-3.
+        **kwargs: Args to pass to the base class.
```
abheesht17, line 17: Is it standard practice in Keras to specify kwargs in the doc-string? Haven't seen it elsewhere.

keras_rs/src/layers/retrieval/sampling_probability_correction.py
```diff
@@ -0,0 +1,41 @@
...
+from keras import ops
+
+from keras_rs.src import types
+from keras_rs.src.api_export import keras_rs_export
+
+
+@keras_rs_export("keras_rs.layers.SamplingProbablityCorrection")
+class SamplingProbablityCorrection(keras.layers.Layer):
+    """Sampling probability correction.
+
```
abheesht17, line 13: Should we write a more descriptive doc-string?

keras_rs/src/layers/retrieval/sampling_probability_correction_test.py
```diff
@@ -0,0 +1,73 @@
...
+from keras_rs.src import testing
+from keras_rs.src.layers.retrieval import sampling_probability_correction
+
+
+class SamplingProbabilityCorrectionTest(testing.TestCase):
+    def test_call(self):
+        shape = (10, 20)  # (num_queries, num_candidates)
+        rng = keras.random.SeedGenerator(42)
+        logits = keras.random.uniform(shape, seed=rng)
+        probs = keras.random.uniform(shape[1:], seed=rng)
```
abheesht17, line 15: We can move these to `setUp()` and reuse them in other testcases

abheesht17 (commented):
Thanks for the PR! Looks good to me. Just some nits

keras_rs/src/layers/retrieval/sampling_probability_correction.py
```diff
@@ -0,0 +1,41 @@
...
+import keras
+from keras import ops
+
+from keras_rs.src import types
+from keras_rs.src.api_export import keras_rs_export
+
+
+@keras_rs_export("keras_rs.layers.SamplingProbablityCorrection")
+class SamplingProbablityCorrection(keras.layers.Layer):
```
abheesht17, line 11: Oh, also, we need a `get_config()` method

keras_rs/src/layers/retrieval/sampling_probability_correction.py
hertschuh: Oh, that's a good question. I don't know if the keras.io scripts handle type annotations.

In code completion, you see the arguments with their types and default values. That's good. Except that `types.Tensor` for instance is replaced by `Any`, because that's how it's defined to not depend on all backends. So maybe for that reason too it makes sense to add the types here.

keras_rs/src/layers/retrieval/sampling_probability_correction.py
hertschuh: Some internal tools flags the docstring as being incorrect if it's not in the args. So I took the habit of always adding it even if it typically does not provide useful information.

keras_rs/src/layers/retrieval/sampling_probability_correction.py
hertschuh: Oh true, I added `epsilon`, which wasn't in the original class, I need a `get_config` now.

keras_rs/src/layers/retrieval/sampling_probability_correction.py
hertschuh: Added.

keras_rs/src/layers/retrieval/sampling_probability_correction.py
hertschuh, line 11: I almost propagated the typo that's been there for years...

abheesht17 (approved):
Thanks!

PR #42, opened 2025-04-07
Use `ml_dtypes` instead of `numpy` for `finfo`.

hertschuh:
This is consistent with Keras' usage of `ml_dtypes`.

Also added `ml_dtypes` as a pip package dependency.


abheesht17 (approved):
Thanks!

Issue #39, opened 2025-03-21
Add support for extra dimensions to `SamplingProbabilityCorrection` and `HardNegativeMining`.


hertschuh:
Same as in `RemoveAccidentalHits`.

PR #43, opened 2025-04-08
Test `SamplingProbabilityCorrection` layer with different dimensions.

hertschuh:
The `SamplingProbabilityCorrection` already supports different dimensions. This adds testing for it.

See https://github.com/keras-team/keras-rs/issues/39


abheesht17 (approved):
Thanks!

PR #44, opened 2025-04-08
Add support for any dimensions to `HardNegativeMining` layer.

hertschuh:
Also simplified the code by using an existing keras op.

Added unit test to cover the case when the inputs have fewer candidates than the requested number of hard negatives.

Fixes https://github.com/keras-team/keras-rs/issues/39


abheesht17 (approved):
Thanks!

PR #46, opened 2025-04-09
Add MRR

abheesht17:
TODO:

- [x] Add UTs.
- [x] Add documentation.
- [x] Verify doc-strings.
- [x] Some small TODOs in the code, which can be decided after code review.


abheesht17:
MRR ready for review

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+                for unbatched inputs or `(batch_size, list_size)` for batched
+                inputs.
+            y_pred: tensor. The predicted values, of shape `(list_size)` for
+                unbatched inputs or `(batch_size, list_size)` for batched
+                inputs. Should be of the same shape as `y_true`.
+            sample_weight: float/tensor. Can be float value, or tensor of
+                shape `(list_size)` or `(batch_size, list_size)`. Defaults to
+                `None`.
+        """
+        # TODO (abheesht): Should `y_true` be a dict, with `"mask"` as one key
```
hertschuh, line 78: You mean as an option? Right now, there's no way to pass a mask, right?

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # === Convert to tensors, if list ===
+        # TODO (abheesht): Figure out if we need to cast tensors to
+        # `self.dtype`.
+        if isinstance(y_true, list):
+            y_true = ops.convert_to_tensor(y_true)
+        if isinstance(y_pred, list):
+            y_pred = ops.convert_to_tensor(y_pred)
+        # `sample_weight` can be a scalar too.
+        if isinstance(sample_weight, (list, float, int)):
```
hertschuh, line 89: Remove all 3 `if isinstance(...)` and just call `ops.convert_to_tensor`, it does the if for you.

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+        # If `sample_weight` rank is 1, it should be of shape `(batch_size,)`.
+        # Otherwise, it should be of shape `(batch_size, list_size)`.
+        if sample_weight_rank == 1:
+            check_shapes_compatible(sample_weight_shape, (y_true_shape[0],))
+            # Uprank this, so that we get per-list weights here.
+            sample_weight = ops.expand_dims(sample_weight, axis=1)
+        elif sample_weight_rank == 2:
+            check_shapes_compatible(sample_weight_shape, y_true_shape)
+
+        # Want to make sure `sample_weight` is of the same shape as
```
hertschuh, line 121: Meaning you should add a check here?

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+        # `y_true`.
+        sample_weight = ops.multiply(ops.ones_like(y_true), sample_weight)
```
hertschuh, line 123: What is this supposed to do? It is to handle the case when `y_true` is ragged / sparse and sample_weight is dense?

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # === Process inputs - shape checking, upranking, etc. ===
```
hertschuh, line 125: Move this comment down 3 lines.

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+        # === Update "invalid" `y_true`, `y_pred` entries based on mask ===
+
+        # `y_true`: assign 0 for invalid entries
+        y_true = ops.where(mask, y_true, ops.zeros_like(y_true))
+        # `y_pred`: assign a value slightly smaller than the smallest value
+        # so that it gets sorted last.
+        y_pred = ops.where(
+            mask,
+            y_pred,
+            -1e-6 * ops.ones_like(y_pred)
```
hertschuh, line 149: Maybe we should use `keras.config.epsilon()` or make it configuration in `__init__`. What do you think?

keras_rs/src/metrics/ranking_metric.py
```diff
@@ -0,0 +1,200 @@
...
+            + ops.amin(y_pred, axis=1, keepdims=True),
+        )
+
+        # === Actual computation ===
+        per_list_metric_values, per_list_metric_weights = self.compute_metric(
+            y_true=y_true, y_pred=y_pred, mask=mask, sample_weight=sample_weight
+        )
+
+        # Chain to `super()` to get mean metric.
+        # TODO (abheesht): Figure out if we want to return unaggregated metric
```
hertschuh, line 159: Return where, right here? When would it be useful?

keras_rs/src/utils/keras_utils.py
```diff
@@ -42,3 +42,16 @@ def check_shapes_compatible(
                 return False
 
     return True
+
+
+# Check ranks and shapes.
+def check_rank(
+    x_rank: int,
+    allowed_ranks: tuple[int, ...] = (1, 2),
+    tensor_name: Optional[str] = None,
```
hertschuh, line 51: It look like `tensor_name` is not really optional.

keras_rs/src/utils/keras_utils.py
```diff
@@ -42,3 +42,16 @@ def check_shapes_compatible(
                 return False
 
     return True
+
+
+# Check ranks and shapes.
+def check_rank(
+    x_rank: int,
+    allowed_ranks: tuple[int, ...] = (1, 2),
```
hertschuh, line 50: I wouldn't provide a default for `allowed_ranks`. When I'm reading the code, I don't want to have to go to the definition to know what ranks I'm checking, I should have them right there where it's called.

keras_rs/src/utils/loss_and_metric_utils.py
```diff
@@ -0,0 +1,62 @@
+from typing import Optional
+
+from keras import ops
+
+from keras_rs.src import types
+from keras_rs.src.utils.keras_utils import check_rank
+from keras_rs.src.utils.keras_utils import check_shapes_compatible
+
+
+def process_inputs(
```
hertschuh, line 10: This have a more specific name, maybe something like `standardize_ranks` or `check_ranks_and_shapes`.

keras_rs/src/utils/loss_and_metric_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    y_true: types.Tensor,
+    y_pred: types.Tensor,
+    mask: Optional[types.Tensor] = None,
+    check_y_true_rank: bool = True,
+) -> tuple[types.Tensor, types.Tensor, Optional[types.Tensor], bool]:
+    """
+    Utility function for processing inputs for losses and metrics.
+
+    This utility function does three things:
```
hertschuh, line 19: Add Args and Return section.

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+        List of sorted tensors (`tensors_to_sort`), sorted using `scores`.
+    """
+    # TODO: Consider exposing `shuffle_ties` to the user.
+    # TODO: Figure out `seed`. How do we propagate it down here from ranking
+    # metric?
+    max_possible_k = ops.shape(scores)[1]
+    if k is None:
+        k = max_possible_k
+    else:
+        k = ops.minimum(k, max_possible_k)
```
hertschuh, line 87: I would split this in 2, because JAX will want a static int value and TensorFlow may need a dynamic value when `max_possible_k` is a scalar tensor:

```python
    if k is None:
        k = max_possible_k
    elif isinstance(max_possible_k, int):
        k = min(k, max_possible_k)
    else:
        k = ops.minimum(k, max_possible_k)
```

hertschuh (commented):
Thanks!

Partial review... to be continued

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+
+    When all the lists have sum(relevance) == 0, we set the average weight to
+    1.0.
+
+    This computation takes care of the following cases:
+    - When all the weights are 1.0, the per-list weights will be 1.0
+      everywhere, even for lists without any relevant examples because
+      `sum(per_list_weights) ==  num(sum(relevance) != 0)`. This handles the
+      standard ranking metrics where the weights are all
+      1.0.
```
hertschuh, line 139: nitpick, move to previous line.

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+      This handles the unbiased metrics well.
+    - For the mixture of the above 2 scenario, the weights for lists with
+      nonzero relevance and nonzero weights is proportional to
+
+      ```
+      per_list_weights / sum(per_list_weights) *
+      num(sum(relevance) != 0) / num(lists)
+      ```
+
+      The rest have weights 1.0 / num(lists).
```
hertschuh, line 150: nit: put formula between single quotes.

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    nonzero_weights = ops.greater(ops.sum(weights, axis=1, keepdims=True), 0.0)
+    # Calculate the sum of relevance per list
+    per_list_relevance = ops.sum(relevance, axis=1, keepdims=True)
+
+    # Identify lists where both weights and relevance sums are non-zero.
+    nonzero_relevance_condition = ops.greater(per_list_relevance, 0.0)
+    nonzero_relevance = ops.where(
+        nonzero_weights,
+        ops.cast(nonzero_relevance_condition, "float32"),
+        ops.zeros_like(per_list_relevance),
```
hertschuh, line 171: Is there a reason to do `zeros_like` rather than just `0.0`? It should give the same result. But is it for ragged tensors or something?

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    # TODO: Pretty nasty function. Check if we can simplify it at a later point.
+
+    # Calculate if the sum of weights per list is greater than 0.0.
+    nonzero_weights = ops.greater(ops.sum(weights, axis=1, keepdims=True), 0.0)
+    # Calculate the sum of relevance per list
+    per_list_relevance = ops.sum(relevance, axis=1, keepdims=True)
+
+    # Identify lists where both weights and relevance sums are non-zero.
+    nonzero_relevance_condition = ops.greater(per_list_relevance, 0.0)
+    nonzero_relevance = ops.where(
```
hertschuh, line 168: It looks like this is just a `ops.logical_and` between `nonzero_relevance` and `nonzero_weights`.

A `logical_and` or even a `multiply` should be faster than a `where`.

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+        ops.cast(nonzero_relevance_condition, "float32"),
+        ops.zeros_like(per_list_relevance),
+    )
+    # Count the number of lists with non-zero relevance (and implicitly
+    # non-zero weights).
+    nonzero_relevance_count = ops.sum(nonzero_relevance, axis=0, keepdims=True)
+
+    # Calculate the per-list weights using the core formula
+    # Numerator: sum(weights * relevance) per list
+    numerator = ops.sum(weights * relevance, axis=1, keepdims=True)
```
hertschuh, line 179: `ops.multiply(weights, relevance)`

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    # relevance > 0)
+    sum_weights = ops.sum(per_list_weights, axis=0, keepdims=True)
+
+    # Calculate the average weight to use as default for lists with zero
+    # relevance but non-zero weights. If no lists have non-zero relevance,
+    # default to 1.0.
+    avg_weight = ops.where(
+        ops.greater(nonzero_relevance_count, 0.0),
+        ops.divide_no_nan(sum_weights, nonzero_relevance_count),
+        ops.ones_like(nonzero_relevance_count),
```
hertschuh, line 193: Just `1.0`?

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    # 2. If sum(weights) > 0 AND sum(relevance) > 0, use the calculated
+    # `per_list_weights`.
+    # 3. If `sum(weights) > 0` AND `sum(relevance) == 0`, use the calculated
+    # `avg_weight`.
+    final_weights = ops.where(
+        nonzero_weights,
+        ops.where(
+            nonzero_relevance_condition,
+            per_list_weights,
+            ops.ones_like(per_list_weights) * avg_weight,
```
hertschuh, line 207: Remove `ones_like`?

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+        ),
+        ops.zeros_like(per_list_weights),
```
hertschuh, line 209: Just `0.0`

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+        ops.ones_like(nonzero_relevance_count),
+    )
+
+    # Final assignment of weights based on conditions:
+    # 1. If sum(weights) == 0 for a list, the final weight is 0.
+    # 2. If sum(weights) > 0 AND sum(relevance) > 0, use the calculated
+    # `per_list_weights`.
+    # 3. If `sum(weights) > 0` AND `sum(relevance) == 0`, use the calculated
+    # `avg_weight`.
+    final_weights = ops.where(
```
hertschuh, line 202: I think it would be faster (and less code) to do

`ops.multiply(ops.cast(nonzero_weights, per_list_weights.dtype), ...)`

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+            ops.ones_like(per_list_weights) * avg_weight,
+        ),
+        ops.zeros_like(per_list_weights),
+    )
+
+    return final_weights
+
+
+def default_gain_fn(label: types.Tensor) -> types.Tensor:
+    return ops.subtract(ops.power(2.0, label), 1.0)
```
hertschuh, line 216: nitpick: I have always assumed `ops.square` is faster.

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def default_rank_discount_fn(rank: types.Tensor) -> types.Tensor:
+    return ops.divide(ops.log(2.0), ops.log1p(rank))
```
hertschuh, line 220: The docstring above says it's equivalent to `lambda rank: log2(rank + 1)`.

So that would be reversing the arguments of the divide. Also, it would be clearer if written as `ops.log2(rank + 1.0)`.

keras_rs/src/metrics/n_dcg.py
```diff
@@ -0,0 +1,154 @@
...
+)
+from keras_rs.src.utils.ranking_metrics_utils import compute_dcg
+from keras_rs.src.utils.ranking_metrics_utils import default_gain_fn
+from keras_rs.src.utils.ranking_metrics_utils import default_rank_discount_fn
+from keras_rs.src.utils.ranking_metrics_utils import get_list_weights
+from keras_rs.src.utils.ranking_metrics_utils import sort_by_scores
+
+
+@keras_rs_export("keras_rs.metrics.nDCG")
+class nDCG(RankingMetric):
```
hertschuh, line 19: The convention is to always use an upper case letter for the first letter of a class, so it should be `NDCG`, even if it's written as `nDCG` in papers.

Also the file should be `ndcg.py`.

I don't actually understand why it's written `nDCG` since all the letters form an acronym anyway.

keras_rs/src/utils/ranking_metrics_utils.py
```diff
@@ -0,0 +1,240 @@
+from typing import Callable, Optional
```
hertschuh, line 1: I'm not a huge fan of the concept of a `utils` folder in general, and I typically prefer the utils files to be in the same folder as where it's used. So for instance, I think this could be in `keras_rs/src/metrics`.

What do you think? Note that keras itself is not consistent on this.

keras_rs/src/utils/pairwise_loss_utils.py
```diff
@@ -1,9 +1,8 @@
-from typing import Callable, Optional
+from typing import Callable
```
hertschuh, line 1: I'm not a huge fan of the concept of a `utils` folder in general, and I typically prefer the utils files to be in the same folder as where it's used. So for instance, I think this could be in `keras_rs/src/losses`.

What do you think? Note that keras itself is not consistent on this.

keras_rs/src/utils/loss_and_metric_utils.py
```diff
@@ -0,0 +1,62 @@
+from typing import Optional
```
hertschuh, line 1: I'm not a huge fan of the concept of a `utils` folder in general, and I typically prefer the utils files to be in the same folder as where it's used. This one is a bit funny because it's for both metrics and losses, but I think it wouldn't be too shocking to put it in `keras_rs/src/metrics` since a loss is just a kind of metric.

What do you think? Note that keras itself is not consistent on this.

hertschuh (commented):
Thanks for this big PR. A lot of work obviously went into this!

More comments...

keras_rs/src/metrics/ranking_metric.py
abheesht17: Right now, we use `sample_weights > 0` to construct the mask.

But for pairwise losses, that doesn't work, and we need to pass mask explicitly. So, should we do the same here?

keras_rs/src/metrics/ranking_metric.py
abheesht17: Yeah, we can use `keras.config.epsilon()`, didn't know it exists :P

keras_rs/src/utils/ranking_metrics_utils.py
abheesht17: Does it typecast to the right dtype if we use 0.0? How does it know from which tensor it should take the dtype?

keras_rs/src/metrics/ranking_metric.py
abheesht17: Nope, the multiplication below takes care of that. I'll amend the comment

keras_rs/src/utils/loss_and_metric_utils.py
abheesht17: Yeah, I agree. I'm not a huge fan of a `utils` folder either, unless it has very generic ops (like checking shape, etc.), which might be used across the repo. I'll move this

keras_rs/src/utils/ranking_metrics_utils.py
hertschuh: Oh you would still need to do `ops.cast(0, per_list_relevance.dtype)`. But hopefully it would use less memory.

keras_rs/src/utils/ranking_metrics_utils.py
abheesht17: It'll get complicated:

```
    zero_weight_mask = ops.cast(ops.logical_not(nonzero_weights), dtype=per_list_weights.dtype)
    nonzero_relevance_mask = ops.cast(nonzero_relevance_condition, dtype=per_list_weights.dtype)
    zero_relevance_mask = ops.cast(ops.logical_not(nonzero_relevance_condition), dtype=per_list_weights.dtype)

    final_weights = (
        zero_weight_mask * 0.0 +  # Case 1: sum(weights) == 0
        nonzero_weights * nonzero_relevance_mask * per_list_weights +  # Case 2: sum(weights) > 0 AND sum(relevance) > 0
        nonzero_weights * zero_relevance_mask * avg_weight  # Case 3: sum(weights) > 0 AND sum(relevance) == 0
    )
```

Every `*` will be replaced by `ops.multiply`. What do you think?

keras_rs/src/utils/ranking_metrics_utils.py
abheesht17: Will address these in a new PR for DCG, nDCG.

keras_rs/src/utils/doc_string_utils.py
```diff
@@ -0,0 +1,48 @@
...
+                textwrap.fill(
+                    stripped_block,
+                    width=width - len(base_indent_str),
+                    initial_indent=base_indent_str,
+                    subsequent_indent=base_indent_str,
+                )
+            )
+
+    final_string = "".join(processed_output).strip()
+    final_string = " " * 4 + final_string
```
hertschuh, line 47: Replace with `base_indent_str`.

keras_rs/src/utils/doc_string_utils.py
```diff
@@ -0,0 +1,48 @@
+import re
+import textwrap
+from typing import Any
+
+
+def format_docstring(template: str, width: int = 80, **kwargs: Any) -> str:
```
hertschuh, line 6: Why is this needed?

abheesht17: The doc-string has "in-line" variables like "concept_sentence". When I do a .format(), the line goes overboard. When I have a newline in concept_sentence, the next line gets messed up

hertschuh (approved):

PR #49, opened 2025-04-13
Add base retrieval class

abheesht17:
Question for you, @hertschuh:

- I have kept doc-strings from `BruteForceRetrieval`'s `call()`, `update_candidates()`, etc., and kept them in the abstract class as well. This leads to duplication, but also ensures complete API documentation on keras.io. Let me know if you want to remove it in one of the classes.
- Is `BaseRetrieval` a better name?


hertschuh (commented):
Thanks for the PR!

To answer your questions:
- I agreed with duplicating the docstrings.
- I prefer `Retrieval` as I see it more as an interface than a base class. The base class does have some validation logic, but all the important methods are abstract.

hertschuh (approved):

Issue #38, opened 2025-03-20
Mismanagement of Public and Private Symbols and Violation of Python's Official Interface Conventions


AlanBogarin:
The current implementation of symbol management in the project relies on the `namex` package, which enforces a structure where the code must reside in src and then creates an api subpackage that references public symbols from src. This approach works partially, but deviates from official conventions for defining public and private interfaces, as outlined in the [Python Typing Specification](https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols).

### Inconsistent Use of `__all__`:
The keras-rs package does not use the `__all__` dunder to define explicitly its public interface in submodules, nor does namex create it or reexport (redundant alias convention).

The `namex` package is only used to expose public symbols by creating another subpackage, but it does not expose the interface properly, according to convention all symbols it exports will be considered private.

The current implementation of the main [`__all__` ](https://github.com/keras-team/keras-rs/blob/7553d191a1a25c39976b0494b439b66477a9c23d/keras_rs/__init__.py#L26) uses an invalid idiom. At runtime, `__all__` is only for restricting star imports, but for type hints, it defines public interfaces, and since these do not execute statements, they cannot determine the current implementation.

### Forced Directory Structure (src and api)

The requirement to place code in src and then generate an api module is an unnecessary abstraction that complicates the codebase and deviates from Python's standard practices.

### Type Checker Compatibility:
The current approach may cause issues with type checkers (e.g., mypy, pyright) that rely on Python's official conventions for determining public and private symbols. This can lead to incorrect type checking and reduced developer productivity.

### Adopt Python's Standard Public/Private Symbol Conventions
Explicitly indicate re-export using the `from module import symbol as symbol` idiom or include it in `__all__` dunder

Fix the public interface `keras_rs.__init__` so that linters can read it

### Remove namex dependency
The namex package introduces unnecessary complexity and does not adhere to Python's official guidelines. Removing this dependency will simplify the codebase and improve compliance with Python's conventions.

### Refactor Directory Structure:
Eliminate the forced src and api structure in favor of a more conventional Python package layout. This will make the project more intuitive for contributors and users.

#### Suggestion
Since the project already creates multiple subpackages with empty `__init__.py` files and many modules, it would be better to define `__all__` within each module and export everything to the `__init__.py` files. This achieves the same goal as `namex` but in a more Pythonic way, avoiding absolute imports in every module and creation of unnecessary modules at runtime.

## Benefits of the Proposed Changes:
[...]

PR #63, opened 2025-04-23
Regenerate APIs with the new version of `namex`.

hertschuh:
Namex was changed to work better with static analyzers and IDE.

Mirrors https://github.com/keras-team/keras-hub/pull/2222 for KerasRS.

Also addresses https://github.com/keras-team/keras-rs/issues/38


mattdangerw (approved):
Thanks!

PR #66, opened 2025-04-27
Add doc-string examples

abheesht17:
Skipped hard neg mining, sample probs, etc.


hertschuh (commented):
Update `namex` and re-run `api_gen`.

I personally have a preference for using Keras ops instead of Numpy in inline examples unless Numpy makes it significantly easier.

- `np.random.randint` -> `keras.random.randint`
- `np.random.randn` -> `keras.random.uniform`
- `np.random.random` -> `keras.random.uniform`

What do you think?

abheesht17:
> Update `namex` and re-run `api_gen`.
> 
> I personally have a preference for using Keras ops instead of Numpy in inline examples unless Numpy makes it significantly easier.
> 
> * `np.random.randint` -> `keras.random.randint`
> * `np.random.randn` -> `keras.random.uniform`
> * `np.random.random` -> `keras.random.uniform`
> 
> What do you think?

We use NumPy everywhere else. Eg: https://github.com/keras-team/keras/blob/v3.9.2/keras/src/losses/losses.py#L639-L640

hertschuh (approved):

PR #72, opened 2025-04-29
Fixes related to documentation rendering.

hertschuh:
- Fix pairwise loss `call` documentation
- Fix example interspersed with numbered list, the numbered list don't work
- Fix comment in metric examples


abheesht17 (approved):
Thanks!

PR #73, opened 2025-04-29
Fix pesky annoying keras.io rendering issues

abheesht17:


keras_rs/src/metrics/dcg.py
```diff
@@ -136,7 +135,7 @@ def from_config(cls, config: dict[str, Any]) -> "DCG":
     ...     y_true=labels, y_pred=scores
     ... )
 
-    # Mask certain elements (can be used for uneven inputs)
+    Mask certain elements (can be used for uneven inputs)
```
hertschuh, line 138: Can you make it consistent with the ones I did yesterday? I.e. add a : and a blank line after.

hertschuh (approved):

PR #81, opened 2025-05-12
Add `Nested` generic type for type hints.

hertschuh:


hertschuh:
> LGTM, but have two questions:
> 
> 1. Where will we use `Nested`?
> 2. Everywhere else, we use `dict` instead of `typing.Mapping`, right? Should we do the same here?  I suppose `Sequence` is okay in the other case, because it subsumes `list`, `tuple`. Or should we use `list`, `tuple` instead of `Sequence` too?

1. Sorry, I should have provided context. This is to support type annotations on arbitrary nested structures, basically all the structures supported by `keras.tree`, which in turn is all the structures supported as inputs and outputs of Keras models. This is because the inputs and outputs of `DistributedEmbedding` will be a `Nested[Tensor]`.

2. Here I use `Mapping` and `Sequence` because we allow other structures than plain `dict`s, like `defaultdict`, `OrderedDict`, anything that `keras.tree` supports. Same for `Sequence`s, we want to support `deque`, `namedtuple` etc.

PR #82, opened 2025-05-12
Add `no_automatic_dependency_tracking` utility.

hertschuh:
This wraps the private equivalent decorators in Keras and TensorFlow.

### DIFF
--- a/keras_rs/src/utils/keras_utils.py
+++ b/keras_rs/src/utils/keras_utils.py
@@ -1,8 +1,30 @@
-from typing import Union
+from typing import Any, Callable, Union
 
 import keras
 
 from keras_rs.src import types
+
+
+def no_automatic_dependency_tracking(
+    fn: Callable[..., Any],
+) -> Callable[..., Any]:
+    """Decorator to disable automatic dependency tracking in Keras and TF.
+
+    Args:
+        fn: the function to disable automatic dependency tracking for.
+
+    Returns:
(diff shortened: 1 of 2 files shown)


keras_rs/src/utils/keras_utils.py
```diff
@@ -1,10 +1,32 @@
...
+    Args:
+        fn: the function to disable automatic dependency tracking for.
+
+    Returns:
+        a wrapped version of `fn`.
+    """
+    if keras.backend.backend() == "tensorflow":
+        import tensorflow as tf
+
+        fn = tf.__internal__.tracking.no_automatic_dependency_tracking(fn)
```
abheesht17, line 22: Quick question, mainly for my understanding: why do we not disable this in core Keras itself?

abheesht17 (approved):

keras_rs/src/utils/keras_utils.py
hertschuh: That is a good question. Apparently, we didn't need it.

Normally, this is about not tracking `Layers`, `Metrics` and `Variables`. The reason I need it is not because of that, it's because dicts, lists, etc. get wrapped into `TrackedDict`, `TrackedList` etc. and the Tensorflow equivalents. I need configuration structures that are nested, but then, when I pass the nested configs to some APIs, that API complains about `TrackedDict`, `TrackedList`, etc. All I want is to preserve the original structures.

PR #87, opened 2025-05-18
Fix embed_reduce for sparse inputs.

cantonios:
Unfortunately `keras.ops.ones_like` discards the sparsity pattern for sparse tensors, instead filling a dense tensor with the same overall shape.

There don't seem to be any standard utilities in Keras for manipulating sparse tensors across multiple backends.  All math functions that operate on sparse tensors intentionally treat explicit `0`s the same as implicit zeros, preventing us from using math tricks to get a sparse tensor of ones.  So, we resort to adding backend-specific checks here.

Currently only TF and JAX are supported.

### DIFF
--- a/keras_rs/src/layers/embedding/embed_reduce.py
+++ b/keras_rs/src/layers/embedding/embed_reduce.py
@@ -10,6 +10,83 @@ from keras_rs.src.utils.keras_utils import check_shapes_compatible
 SUPPORTED_COMBINERS = ("mean", "sum", "sqrtn")
 
 
+def _is_supported_sparse(x: types.Tensor) -> bool:
+    """Determines if the input is a supported sparse tensor.
+
+    NOTE: Currently only works for the TensorFlow and JAX backends.
+
+    Args:
+      x: Input tensor to check for sparsity.
+
+    Returns:
+      True if `x` is a supported sparse tensor.
+    """
+    if keras.backend.backend() == "tensorflow":
+        import tensorflow as tf
+
+        return isinstance(x, tf.SparseTensor)
+    elif keras.backend.backend() == "jax":
+        from jax.experimental import sparse as jax_sparse
+
+        return isinstance(x, jax_sparse.BCOO) or isinstance(x, jax_sparse.BCSR)
+
+    return False
+
+
+def _sparse_ones_like(
+    x: types.Tensor, dtype: Optional[types.DType] = None
+) -> types.Tensor:
+    """Creates a tensor of ones with the same sparsity as the input.
+
+    This differs from `keras.ops.ones_like`, which would create a dense
+    tensor of ones.
+
+    Args:
+        x: Input sparse tensor.
+        dtype: Optional dtype for the output tensor values.
+
+    Returns:
+        Sparse tensor of ones.
+
+    Raises:
+        ValueError for unsupported sparse input type and backend.
+    """
+    dtype = dtype or x.dtype
+    if keras.backend.backend() == "tensorflow":
+        import tensorflow as tf
+
+        # Ensure shape is copied exactly for compatibility in graph mode.
+        x_shape = x.shape
+        y = tf.SparseTensor(
+            x.indices, tf.ones_like(x.values, dtype=dtype), x.dense_shape
+        )
+        y.set_shape(x_shape)
+        return y
+    elif keras.backend.backend() == "jax":
+        import jax.numpy as jnp
+        from jax.experimental import sparse as jax_sparse
+
+        if isinstance(x, jax_sparse.BCOO):
+            return jax_sparse.BCOO(
+                (jnp.ones_like(x.data, dtype=dtype), x.indices),
+                shape=x.shape,
+                indices_sorted=x.indices_sorted,
+                unique_indices=x.unique_indices,
+            )
+        elif isinstance(x, jax_sparse.BCSR):
+            return jax_sparse.BCSR(
+                (jnp.ones_like(x.data, dtype=dtype), x.indices, x.indptr),
+                shape=x.shape,
+                indices_sorted=x.indices_sorted,
+                unique_indices=x.unique_indices,
+            )
+
+    raise ValueError(
+        f"Unsupported sparse input type '{x.__class__.__name__}' for backend "
+        f"{keras.backend.backend()}."
+    )
+
+
 @keras_rs_export("keras_rs.layers.EmbedReduce")
 class EmbedReduce(keras.layers.Embedding):
     """An embedding layer that reduces with a combiner.
@@ -128,7 +205,7 @@ class EmbedReduce(keras.layers.Embedding):
         #   number of items per row.
         # - For sparse inputs, after embedding, we get a dense tensor, not a
         #   sparse tensor. What it does for missing values is use embedding 0.
-        #   These are bogus embedding ands should be ignored. `ones_like` gives
+        #   These are bogus embedding and should be ignored. `ones_like` gives
         #   us a sparse tensor with the exact same missing values. Later, when
         #   we do `x = ops.multiply(x, weights)`, which masks the bogus values
         #   (note that `weights` has been densified beforehand). Additionally,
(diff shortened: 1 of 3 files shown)


hertschuh:
Oh, that's fixed in Keras core, you just have to use Keras-nightly, which is what this package is using right now. We'll release Keras on Monday.

keras_rs/src/layers/embedding/embed_reduce.py
```diff
@@ -10,6 +10,77 @@
...
+        Sparse tensor of ones.
+
+    Raises:
+        ValueError for unsupported sparse input type and backend.
+    """
+    dtype = dtype or x.dtype
+    if keras.backend.backend() == "tensorflow":
+        import tensorflow as tf
+
+        return tf.sparse.map_values(tf.ones_like, x, dtype=dtype)
```
hertschuh, line 58: Unfortunately, `tf.sparse.map_values` loses the static shape, which almost always causes issues in graph mode. So I created this as a workaround:
https://github.com/keras-team/keras/blob/master/keras/src/backend/tensorflow/sparse.py#L25
```
sparse_with_values(x, tf.ones_like(x.values, dtype=dtype))
```

keras_rs/src/layers/embedding/embed_reduce.py
```diff
@@ -10,6 +10,77 @@
...
+    elif keras.backend.backend() == "jax":
+        from jax.experimental import sparse as jax_sparse
+        import jax.numpy as jnp
+
+        if isinstance(x, jax_sparse.BCOO):
+            return jax_sparse.BCOO(
```
hertschuh, line 64: There is a subtlety here, it could return the wrong result.

JAX allows itself to be lazy and keep duplicate indices (in part to allow compilations). But then it can create multiple ones for duplicate indices which eventually get added later. So you can end up with twos, threes... etc. So for non-linear ops, you have to dedupe beforehand.

See here: https://github.com/keras-team/keras/blob/master/keras/src/backend/jax/sparse.py#L150-L151

hertschuh (changes requested):

keras_rs/src/layers/embedding/embed_reduce.py
cantonios: Interesting.  That sounds like a TF bug, but maybe it's because the `map_values` function may involve multiple sparse tensors?  Anyway, I'll copy your approach until those sparse helper functions are exposed in the public API.

keras_rs/src/layers/embedding/embed_reduce.py
cantonios: In our particular application there shouldn't be duplicates.  The row represents the feature sample row within the batch, and the column is the `position` of the token within that feature sample.  If there was a duplicate `(row, col)`, it means they've messed up constructing the sample itself.  Duplicate tokens (the `values` of the sparse matrix) are fine.  If there _was_ a duplicate `(row, col)` entry, and if you de-dupped them, that would end up summing the token values, resulting in a completely different (and incorrect) token.

cantonios: Do you know what the "formatting" failure is:
```
AttributeError: 'MessageFactory' object has no attribute 'GetPrototype'
```
?

keras_rs/src/layers/embedding/embed_reduce.py
hertschuh: Yes it is, basically all TF sparse ops have this bug. I fixed one here: https://github.com/tensorflow/tensorflow/commit/8056627d6a794f9acaff1a8165cc247bd252db31

Then I decided to fix them in keras instead.

keras_rs/src/layers/embedding/embed_reduce.py
hertschuh: > Do you know what the "formatting" failure is:
> 
> ```
> AttributeError: 'MessageFactory' object has no attribute 'GetPrototype'
> ```
> 
> ?

This is from `ruff`? Or `mypy`? I've never seen this.

cantonios: > `EmbedReduce` is a general purpose layer that is exposed publicly (I had people asking for it who are going to adopt it). There is no assumption that can be made on the indices and the duplicates. There is no garantee that they come from the dataset directly.
> 
> if you do this (`x` is sparse, `y` is sparse), you will have duplicate indices:
> 
> ```
> z = keras.ops.concatenate((x, y), axis=1)
> e = my_reduce_layer(z)
> ```
> 
> In Keras I added unit tests to cover this since is creates bugs that are really confusing.

By "our application", I meant any embed reduce application in general.  Duplicate positions in the `x` (input IDs) tensor will lead to incorrect results regardless.

Consider a sparse `inputs` tensor:
```
(0, 0) = 2
(0, 0) = 3
(0, 1) = 5
```
The `(row, col)` are positions within the `inputs` tensor of embedding IDs.  The `{2, 3, 5}` are the embedding IDs themselves, which for LLM inputs might be word tokens.  The token `2` might correspond to the word `"hello"`, the token `3` to `"beautiful"`, and `5` to `"world"`.  The original input read `{"hello", "beautiful", "world"}`. If you do a `sum_duplicates`, you would get:
```
(0, 0) = 5
(0, 1) = 5
```
Now you have a completely different input tensor `{"world", "world"}`.

You could even lead to a crash.  Consider another example with a vocabulary size of 10:
```
(0, 0) = 5
(0, 0) = 6
(0, 0) = 7
```
Doing a `sum_duplicates` results in
```
(0, 0) = 18
```
This is outside the vocabulary and doesn't even have embedding values to `take`.  If there are any duplicate positions in the sparse input IDs tensor, it indicates a bug, regardless of whether they come directly from a dataset or an intermediate layer generating the IDs.

Digging through, it looks like `keras.op.take(...)` (and hence `keras.layers.embedding`) will end up summing duplicate entries for sparse inputs, but only because we use the generic `convert_to_tensor(x, sparse=False)`, which densifies the input.  I can dedup for consistency here, but I still say it's a bug.

cantonios: > > Do you know what the "formatting" failure is:
> > ```
> > AttributeError: 'MessageFactory' object has no attribute 'GetPrototype'
> > ```
> >   
> > ?
> 
> This is from `ruff`? Or `mypy`? I've never seen this.

This is in the failing check for this PR.

cantonios: Last update: we can't sum duplicates in JAX because:

> ```Additionally, this operation will result in explicit zero entries removed ...```

so doing so would remove any elements corresponding to embedding ID 0.

hertschuh: > Update: ah, TF crashes if there are duplicate indices when converting to dense. Again, this to me indicates that duplicate indices for embedding IDs is a bug.

But TF should never add duplicate indices itself, right? So as long as the initial inputs are correct, this should never happen. I didn't realize that it would allow it in the constructor.

hertschuh: > > > Do you know what the "formatting" failure is:
> > > ```
> > > AttributeError: 'MessageFactory' object has no attribute 'GetPrototype'
> > > ```
> > 
> > This is from `ruff`? Or `mypy`? I've never seen this.
> 
> This is in the failing check for this PR.

It looks like it's `api_gen`. My guess is that this indicates a version of protobuf that doesn't play nice with the version of tensorflow. But why?

hertschuh: As part of this PR, try to revert the changes I made to `requirements.txt` and `requirements-common.txt`, that might help:

https://github.com/keras-team/keras-rs/pull/83/files#diff-fe5b0ea82c3318f9ade681599730162c1df70d8a638eb399ae08210046e3f22a

cantonios: > True. let's just throw an error if unique_indices is False.

The problem with that is the value `unique_indices` defaults to `False`, and doesn't necessarily reflect whether or not there are duplicated indices, only that nobody set it to `True` (either in the constructor, or by calling something like `.sum_duplicates()`).

We can potentially write our own check for duplicated indices, but it will involve a sort computation and runtime check on tensor values.

The only other approach I can think of is to create our own version of `sum_duplicates` that doesn't remove explicit zeros.  Or perhaps create an `overwrite_duplicates` that will keep the last entry, which is probably more correct that summing when dealing with a tensor of indices.

> But TF should never add duplicate indices itself, right?

Correct.  Does JAX ever do this?

hertschuh: Yes, for instance `add` is implemented purely as concat indices and concat values.

The actual addition of values and deduping of indices only happens if and when you do `.sum_duplicates()`.

hertschuh (approved):

PR #90, opened 2025-05-20
Add backend/hardware-independent doc for `DistributedEmbedding`.

hertschuh:
Specific TF and JAX documentation will be added with support.

Also:
- fixed `EmbedReduce` doc rendering
- simplified `TableConfig` default initializer by removing parameter that is the default

### DIFF
--- a/keras_rs/src/layers/embedding/base_distributed_embedding.py
+++ b/keras_rs/src/layers/embedding/base_distributed_embedding.py
@@ -24,6 +24,74 @@ PlacementAndPath = collections.namedtuple(
 class DistributedEmbedding(keras.layers.Layer):
     """DistributedEmbedding, a layer for accelerated large embedding lookups.
 
+    ---
+
+    ## Note: `DistributedEmbedding` is in Preview.
+
+    ---
+
+    ## Configuration
+
+    A `DistributedEmbedding` embedding layer is configured via a set of
+    `keras_rs.layers.FeatureConfig` objects, which themselves refer to
+    `keras_rs.layers.TableConfig` objects.
+
+    - `TableConfig` defines an embedding table with parameters such as its
+      vocabulary size, embedding dimension, as well as a combiner for reduction
+      and optimizer for training.
+    - `FeatureConfig` defines what input features the `DistributedEmbedding`
+      will handle and which embedding table to use. Note that multiple features
+      can use the same embedding table.
+
+    ```python
+    table1 = keras_rs.layers.TableConfig(
+        name="table1",
+        vocabulary_size=TABLE1_VOCABULARY_SIZE,
+        embedding_dim=TABLE1_EMBEDDING_SIZE,
+    )
+    table2 = keras_rs.layers.TableConfig(
+        name="table2",
+        vocabulary_size=TABLE2_VOCABULARY_SIZE,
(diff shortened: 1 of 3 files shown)


keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -24,6 +24,76 @@
...
+    ## Optimizers
+
+    Each embedding table within `DistributedEmbedding` uses its own optimizer
+    for training, which is independent from the optimizer set on the model via
+    `model.compile()`.
+
+    Note that not all optimizers are supported. Currently, the following are
+    supported:
+
+    - `keras.optimizers.Adagrad`
```
cantonios, line 88: Note that JAX only currently supports Adagrad and SGD.  We default to SGD if you specify a different one.
I don't raise an error because the default "Adam" isn't supported.

keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -200,6 +270,24 @@ def call(
         weights: Optional[types.Nested[types.Tensor]] = None,
         training: bool = False,
     ) -> types.Nested[types.Tensor]:
+        """Lookup features in embedding tables and apply reduction.
+
+        Args:
+            inputs: A nested structure of 1D tensors to embed, or 2D tensor to
```
cantonios, line 276: nit: 2D tensor**s**.

cantonios (approved):

keras_rs/src/layers/embedding/base_distributed_embedding.py
hertschuh: Will fix, but what is supported on JAX for dense tensors? On TF, they have to either be 1D or be `(batch_size, 1)`.

hertschuh: Meaning that with dense tensors you can't actually do any reduction.

cantonios: `(batch_size, N)`.  It can take a full dense tensor, with reduction.  Same as if it were ragged with each row the same width.

hertschuh: That makes sense, I'll try to add some pre/post processing in `call` to support that on TF.

keras_rs/src/layers/embedding/base_distributed_embedding.py
hertschuh: Oh, should we change the default in `TableConfig` then?

cantonios: You could do.  I don't really mind either way.  The optimizers I think usually default to Adam in keras.

hertschuh: Will JAX support Adam soonish?

cantonios: Depends on the priority and what "soonish" means.  It's on the TODO list.

hertschuh: Ok, I'm leaving "Adam" as the default then.

PR #94, opened 2025-05-21
Added TF specific documentation to `DistributedEmbedding`.

hertschuh:


keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -30,6 +30,19 @@ class DistributedEmbedding(keras.layers.Layer):
 
     ---
 
+    `DistributedEmbedding` is a layer optimized for TPU chips with SparseCore
+    and can dramatically improve the speed of embedding lookups and embedding
+    training, in particular for large embedding tables. It works by combining
+    multiple lookups in one invocations, and by sharding the embedding tables
```
cantonios, line 36: multiple lookups into one invocation,

keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -30,6 +30,19 @@ class DistributedEmbedding(keras.layers.Layer):
...
+    across the available chips.
```
cantonios, line 37: Sharding doesn't necessarily make it faster on its own.  That leads to more communication for the collective, which can actually slow the lookup down if the table would otherwise fit on a single chip.  We might want to specify that it can make lookups faster for large tables that already need to be model-sharded.

keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -83,15 +96,96 @@ class DistributedEmbedding(keras.layers.Layer):
...
+
+    ### Using in a Keras model
+
+    The use Keras' `model.fit()`, one must compile the model under the
+    `TPUStrategy`. Then, `model.fit()`, `model.evaluate()` or `model.predict()`
+    can be called directly. The Keras model takes care of running the model
+    using the strategy and also automatically distributes the dataset.
+
+    ```python
+    model = create_model(embedding)
```
cantonios, line 143: Should you create the model under the scope as well, since the layer needs to be?

cantonios (changes requested):

keras_rs/src/layers/embedding/base_distributed_embedding.py
hertschuh: Changed wording and added a whole section about placement.

keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -32,7 +32,25 @@ class DistributedEmbedding(keras.layers.Layer):
 
     ---
 
-    ## Configuration
+    `DistributedEmbedding` is a layer optimized for TPU chips with SparseCore
+    and can dramatically improve the speed of embedding lookups and embedding
+    training. It works by combining multiple lookups into one invocations, and
```
cantonios, line 37: invocations -> invocation

keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -78,22 +98,141 @@ class DistributedEmbedding(keras.layers.Layer):
...
+      typically be replicated across TPUs by using the
+      `keras.distribution.DataParallel` distribution option.
+
+    ### Usage with TensorFlow on TPU with SpareCore
+
+    #### Inputs
+
+    In addition to `tf.Tensor`, `DistributedEmbedding` accepts `tf.RaggedTensor`
+    and `tf.SparseTensor` as inputs for the embedding lookups. Ragged tensors
+    must be ragged in dimension 1. Note that if weights are passed, each weight
```
cantonios, line 156: Is dimension 1 the dimension with index 0?  Does TF actually support any other kind of raggedness?

cantonios (approved):
Approval with a couple minor nits in-line.

keras_rs/src/layers/embedding/base_distributed_embedding.py
hertschuh: I meant dimension with index 1. TF supports any ragged dimension and in fact multiple ragged dimensions, a tensor can be twice ragged.

https://www.tensorflow.org/api_docs/python/tf/RaggedTensor#multiple_ragged_dimensions

https://www.tensorflow.org/api_docs/python/tf/RaggedTensor#attributes (see `ragged_rank`)

PR #100, opened 2025-05-22
Remove dependency on `tree` and use `keras.tree`.

hertschuh:
Keras can already depend on either `dmtree` or `optree` and use whichever is best or available on the current platform.


cantonios (approved):
Though note that jax-tpu-embedding depends on `tree` internally, so it will still be pulled in as a dependency.

PR #107, opened 2025-05-23
Add ragged support for default_device placement on JAX.

cantonios:
Requires calling `preprocess`.

Internally, we currently convert ragged inputs to dense before passing to the embedding call(...) function.

### DIFF
--- a/keras_rs/src/layers/embedding/base_distributed_embedding.py
+++ b/keras_rs/src/layers/embedding/base_distributed_embedding.py
@@ -3,6 +3,7 @@ import typing
 from typing import Any, Sequence
 
 import keras
+import numpy as np
 from keras.src import backend
 
 from keras_rs.src import types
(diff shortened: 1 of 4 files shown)


keras_rs/src/layers/embedding/base_distributed_embedding.py
```diff
@@ -655,8 +713,40 @@ def _default_device_preprocess(
         inputs: dict[str, types.Tensor],
         weights: dict[str, types.Tensor] | None,
         training: bool = False,
-    ) -> dict[str, types.Tensor]:
+    ) -> dict[str, dict[str, types.Tensor]]:
         del training
+
+        if keras.backend.backend() == "jax":
```
hertschuh, line 719: Can you add a comment here explaining why we do `if jax` here instead of this being an override in the JAX specific subclass.

PR #112, opened 2025-05-27
Fix tensor shape mismatch by squeezing user_gender_X_raw_user_age input

kevinnkansah:
The user_gender_X_raw_user_age feature has an extra dimension compared to the other features.


hertschuh:
Fixing here: https://github.com/keras-team/keras-io/pull/2114

PR #113, opened 2025-05-29
Enable preprocess calls with symbolic input tensors.

cantonios:
This allows us to more-easily create functional models via:
```python
preprocessed_inputs = distributed_embedding.preprocess(symbolic_inputs, symbolic_weights)
outputs = distributed_embedding(preprocessed_inputs)
model = keras.Model(inputs=preprocessed_inputs, outputs=outputs)
```

Replaced the `keras.Sequential` model in the tests to a functional model to be consistent with TF.

Some other minor changes to address warnings, specifically:
- Explicitly set `steps_per_epoch=1` to avoid warnings about running out of data when using a generator
- Replace `jnp.concatenate` with `np.concatenate` in preprocessing to avoid mixing tensors and non-tensors


keras_rs/src/layers/embedding/distributed_embedding_test.py
```diff
@@ -318,6 +319,37 @@ def test_model_fit(self, input_type, use_weights):
         with self._strategy.scope():
             layer = distributed_embedding.DistributedEmbedding(feature_configs)
 
+        # Mark ragged inputs as sparse in JAX for the purposes
+        # of preprocessing.  Trying to create a "ragged=True" keras.Input for
```
hertschuh, line 323: Well, should we just change that?

keras_rs/src/layers/embedding/distributed_embedding_test.py
```diff
@@ -318,6 +319,37 @@ def test_model_fit(self, input_type, use_weights):
         with self._strategy.scope():
             layer = distributed_embedding.DistributedEmbedding(feature_configs)
 
+        # Mark ragged inputs as sparse in JAX for the purposes
```
hertschuh, line 322: Hmm... I don't think it's a good idea, what if you don't mark it sparse either? Does it work?

keras_rs/src/layers/embedding/distributed_embedding_test.py
cantonios: I thought about it - Keras throws an exception for good reason: the backend is marked as not supporting Ragged.  It does work if I remove that check from `keras.KerasTensor`.

keras_rs/src/layers/embedding/distributed_embedding_test.py
cantonios: I had reservations about this too, and don't really want users to do the same thing.

If you don't mark it as sparse you get warnings about the actual model inputs not matching the original functional model definition.

For ragged inputs and no explicit weights, when we densify we need to create a new weights tensor to mask out the padding.  We end up in a situation where the functional model expects inputs without weights, but after `preprocess`, we do have a weights tensor.

If you think it's okay to remove the check from `keras.KerasTensor`, then we only use the `ragged` attribute here, that's probably best from a user perspective.

hertschuh: > If you don't mark it as sparse you get warnings about the actual model inputs not matching the original functional model definition.

What's the warning?

> We end up in a situation where the functional model expects inputs without weights, but after preprocess, we do have a weights tensor.

I'm not following how that is related to ragged vs. not.

> If you think it's okay to remove the check from keras.KerasTensor, then we only use the ragged attribute here, that's probably best from a user perspective.

Which check? I think we shouldn't change the check in `KerasTensor` but change jax's `SUPPORTS_RAGGED_TENSORS` to True.

keras_rs/src/layers/embedding/distributed_embedding_test.py
```diff
@@ -318,6 +320,32 @@ def test_model_fit(self, input_type, use_weights):
...
+            # JAX doesn't allow us to directly mark an input as ragged, since
+            # Keras+JAX doesn't directly support ragged tensors.
+            # To work around this, explicitly modify the attribute after creation.
+            tensor_input = keras.layers.Input(
+                feature_config.input_shape[1:],
+                dtype=dtype,
+                sparse=(input_type == "sparse"),
+                ragged=False,
+            )
+            tensor_input._ragged = input_type == "ragged"
```
cantonios, line 335: How about this as a work-around?

This let's us keep Keras as-is (not actually supporting ragged tensors), but we can still "detect" a ragged symbolic tensor to avoid the warning from the functional model.

hertschuh: Yeah, I suppose that's ok too. Up to you.

cantonios: Alright, marking as `sparse`.  PTAL.

PR #122, opened 2025-06-09
Checkpoints

silkyarora:


keras_rs/src/layers/embedding/jax/checkpoint_utils.py
```diff
@@ -0,0 +1,103 @@
...
+        self._checkpoint_dir = checkpoint_dir
+        super().__init__(checkpoint_dir, options=options)
+
+    def _get_state(self) -> tuple[dict[str, Any], dict[str, Any]]:
+        """Gets the model state and metrics"""
+        model_state = self._model.get_state_tree()
+        state = {}
+        metrics = {}
+        for k, v in model_state.items():
+            if k == "metrics_variables":
```
hertschuh, line 35: Any particular reason to exclude metrics?

keras_rs/src/layers/embedding/jax/checkpoint_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            restores the latest step.
+        """
+        if step is None:
+            step = self.latest_step()
+        # Restore the model state only, not metrics.
+        state, _ = self._get_state()
+        restored_state = self.restore(
+            step, args=ocp.args.StandardRestore(item=state)
+        )
+        self._model.set_state_tree(restored_state)
```
hertschuh, line 67: This works even without passing the metrics?

keras_rs/src/layers/embedding/jax/distributed_embedding.py
```diff
@@ -384,7 +384,7 @@ def sparsecore_build(
     ) -> None:
         del input_shapes  # Unused.
 
-        if self._sparsecore_built:
+        if hasattr(self, "_sparsecore_built") and self._sparsecore_built:
```
hertschuh, line 387: nitpick: `getattr(self, "_sparsecore_built", False)` is a bit shorter

keras_rs/src/layers/embedding/jax/distributed_embedding_test.py
```diff
@@ -458,6 +464,127 @@ def loss_fn(y_true, y_pred):
...
+        feature_configs_with_zero_init = {
+            feature_config.name: dataclasses.replace(
+                feature_config,
+                table=dataclasses.replace(
+                    feature_config.table, initializer="zeros"
+                ),
+            )
+            for feature_config in feature_configs
+        }
+        layer_2 = jax_distributed_embedding.DistributedEmbedding(
```
hertschuh, line 562: nitpick: rename `layer_for_restore` for consistency with the other variables below

hertschuh (commented):
Thanks for the PR!

keras_rs/src/layers/embedding/jax/checkpoint_utils.py
silkyarora: In orbax metrics can be traced separately. I added those in save method.

keras_rs/src/layers/embedding/jax/checkpoint_utils.py
silkyarora: Yes, restore only restores the current state tree as passed in the StandardRestore Args.

keras_rs/src/layers/embedding/jax/distributed_embedding.py
silkyarora: Done.

keras_rs/src/layers/embedding/jax/distributed_embedding.py
```diff
@@ -384,7 +384,7 @@ def sparsecore_build(
     ) -> None:
         del input_shapes  # Unused.
 
-        if self._sparsecore_built:
+        if getattr(self, "_sparsecore_built", False):
```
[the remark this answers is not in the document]
hertschuh: That's a good point. Why was this change needed in the first place?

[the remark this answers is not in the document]
silkyarora: It is not a required change. While debugging I ran into the not build state for sparsecore and got attribute not found exception.  Looking further I did find that this field is only set in `__sparsecore_init` and added a fix to not run into an exception. 

I am okay to undo this change.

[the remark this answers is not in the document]
cantonios: Yes, please.  My guess is this came up when trying to explicitly call `__sparsecore_build` when no sparsecores were available.

PR #123, opened 2025-06-17
Add FTRL and Adam optimizers.

cantonios:
Also added checks for unsupported options.


cantonios:
Test failure is a dependency issue with upstream jax-tpu-embedding package.  Fixing that over there.

hertschuh (approved):

PR #124, opened 2025-06-18
Update references of jax_layout.Layout to jax_layout.Format.

cantonios:
This reflects an upstream change in JAX.

### DIFF
--- a/requirements-jax-cuda.txt
+++ b/requirements-jax-cuda.txt
@@ -8,7 +8,7 @@ torch>=2.1.0
 # Jax with cuda support.
 # Keep same version as Keras repo.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda12_pip]==0.6.0
+jax[cuda12_pip]==0.6.2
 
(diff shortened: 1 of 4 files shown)


requirements-jax-cuda.txt
```diff
@@ -8,7 +8,7 @@ torch>=2.1.0
 # Jax with cuda support.
 # Keep same version as Keras repo.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda12_pip]==0.6.0
+jax[cuda12_pip]==0.6.2
```
hertschuh, line 11: How is this even working?

https://github.com/jax-ml/jax/blob/008522175ec75f4c2d70f6a6ea56e0d2da9771ed/CHANGELOG.md?plain=1#L69

cantonios: Interesting.  Maybe it's not working?  I didn't actually try this.

hertschuh: Yeah, it's never been used I believe, so we wouldn't know if it's broken.

PR #133, opened 2025-07-25
Fix JAX Layout tiling argument for JAX 0.7.1

hertschuh:
The `jax.experimental.layout.Layout` argument is renamed from `_tiling` to `tiling`.


abheesht17 (approved):
Thanks!

Issue #134, opened 2025-07-29
NaN loss when some embedding tables are on SC, and the rest are on default device


abheesht17:
Reproduced here: https://colab.research.google.com/gist/abheesht17/0fd87eba48b18ad926428df177ab386c/dist-embedding-bug-mismatch-in-features-when-both-default-device-sc-are-used.ipynb.

Did some basic debugging. The problem is that the order of input features gets messed up. For example, assume we have multiple embedding tables, with the first two on SC, the third on `default_device`.

From the dataloader:
```
orig [feature_1] ---> tf.Tensor(
[[20115518  6197194 32254581]
 [37204998  3374086 35559788]
 [ 2813870 11701359  2189495]
 [24173004  4887073 14568268]
 [26047562  8705109 32734126]
 [ 6200587 39533992 22569755]
 [17820371 15671598   823892]
 [33170636 24803444 30541610]], shape=(8, 3), dtype=int64)
```

Inside `base_distributed_embedding`:

```
orig [feature_3] ---> tf.Tensor(
[[20115518  6197194 32254581]
 [37204998  3374086 35559788]
 [ 2813870 11701359  2189495]
 [24173004  4887073 14568268]
 [26047562  8705109 32734126]
 [ 6200587 39533992 22569755]
 [17820371 15671598   823892]
 [33170636 24803444 30541610]], shape=(8, 3), dtype=int64)
```

Clearly, `feature_3` is taking values from `feature_1`.

abheesht17:
Opening a PR in a bit.

abheesht17:
https://github.com/keras-team/keras-rs/blob/main/keras_rs/src/layers/embedding/base_distributed_embedding.py#L657-L663

PR #135, opened 2025-07-29
Fix distributed embedding processing

abheesht17:
Resolves https://github.com/keras-team/keras-rs/issues/134


hertschuh (commented):
Thanks for discovering and investigating this bug!

I think it's missing the reverse index mapping for the outputs in `call`.

But I'm proposing https://github.com/keras-team/keras-rs/pull/136 instead.

It doesn't use an extra data structure for the indices, instead it uses the existing `self._placement_to_path_to_feature_config`.

PR #136, opened 2025-07-30
Fix handling of mixed placement in `DistributedEmbedding`.

hertschuh:
The transformation from the deeply nested input structure to the intermediate representation of "placement -> path -> tensor" and the reverse transformation for the outputs were incorrectly based on flattening and packing as `self._placement_to_path_to_feature_config`. This would use an incorrect order for the flat structure and result in tensors at the wrong place.

This fix makes use of `self._placement_to_path_to_feature_config` instead, which can correctly do the mapping.

Also changed `PlacementAndPath` to be a `dataclass` so that `keras.tree` considers those atoms and doesn't recurse into them. This simplifies some of the `map_structure{_up_to}` calls.

Fixes https://github.com/keras-team/keras-rs/issues/134


keras_rs/src/layers/embedding/distributed_embedding_test.py
```diff
@@ -685,6 +685,77 @@ def test_shared_table(self):
...
+            vocabulary_size=VOCABULARY_SIZE,
+            embedding_dim=embedding_output_dim3,
+            placement="default_device",
+        )
+
+        embedding_config = {
+            "feature1": config.FeatureConfig(
+                name="feature1",
+                table=table1,
+                input_shape=(BATCH_SIZE_PER_CORE, 1),
```
abheesht17, line 722: Why do we call it `BATCH_SIZE_PER_CORE`? Thought we changed the documentation to using `GLOBAL_BATCH_SIZE`?

hertschuh: Oh, this is incorrect indeed. Will fix everywhere (not just this test) in a separate PR.

PR #139, opened 2025-08-05
Fix for batch size of `FeatureConfig.output_shape` on TensorFlow.

hertschuh:
The TensorFlow's `FeatureConfig.output_shape` expects the batch size to be per replica. However, `keras_rs` use the global batch size in `input_shape` and `output_shape`. This modifies the conversion code to take this into account.


abheesht17 (approved):
Nice, LGTM!

Issue #143, opened 2025-08-17
Multi-host training does not work for distributed embeddings


abheesht17:
Code (sourced from [this example](https://keras.io/keras_rs/examples/distributed_embedding_jax/)): https://gist.github.com/abheesht17/0491a8d34756a208a000ff867ad7cc13

This is the error which crops up: https://gist.github.com/abheesht17/694a07e33d217b23692105e281a91fe4

This fixed the forward pass, but I don't know if this is the correct way to do it. Instead of sharding the dataset using Keras' `distribution.distribute_dataset`, shard it manually using JAX (before sharding): https://github.com/AI-Hypercomputer/RecML/blob/dlrm_mlperf/examples/dlrm/dlrm_main.py#L347-L357. However, `fit()` fails because it tries to shard the dataset again.

abheesht17:
@cantonios - I know you're aware of this issue, but created one just to keep track. Thanks!

PR #144, opened 2025-08-18
[MLPerf] Add DLRM-DCNv2

abheesht17:
TODOs:

- [x] Need to make it work for multi-host: https://github.com/keras-team/keras-rs/issues/143.
- [x]  Currently uses dummy data, need to shift it to using the actual dataset. Already added code for loading actual data, but the path we are reading already has "batched" data. Can go ahead with this, or can load "unbatched" data and batch that up.
- [x] Eval dataset
- [x] Check if random seed is being correctly passed. We need to make sure that the initialiser returns different values for different layers (since we are passing `SeedGenerator`, I think we are good, but good to verify manually).
- [ ] ~Instead of separate files for configs, maybe have separate functions for configs.~
- [ ] Is this a worry?

```
 08:44:12.046114: W jax_tpu_embedding/sparsecore/lib/core/input_preprocessing_util.cc:251] No Coo Buffer Size provided for table cat_14_table_cat_15_table_cat_23_table_cat_24_table_cat_25_table_cat_33_table_cat_34_table_cat_35_table_cat_36_table, the default value (6144) may be too large and can cause OOM. Utilize the stats returned from the sparse dense matmul preprocessing API.
```
- [ ] Set up metric logging and checkpointing.


examples/ml_perf/main.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+        vocabulary_size = large_emb_feature["vocabulary_size"]
+        multi_hot_size = large_emb_feature["multi_hot_size"]
+
+        table_config = keras_rs.layers.TableConfig(
+            name=f"{feature_name}_table",
+            vocabulary_size=vocabulary_size,
+            embedding_dim=embedding_dim,
+            # TODO(abheesht): Verify.
+            initializer=keras.initializers.VarianceScaling(
```
hertschuh, line 57: The default is `keras.initializers.VarianceScaling(mode="fan_out")`, which has a scale of 1 and a uniform distribution.

Why did you need to change it?

examples/ml_perf/main.py
```diff
@@ -0,0 +1,242 @@
...
+os.environ["KERAS_BACKEND"] = "jax"
+
+import keras
+
+import keras_rs
+
+from .dataloader import DataLoader
+from .model import DLRMDCNV2
+
+SEED = 1337
```
hertschuh, line 14: It looks like you carefully passed the SEED everywhere, but it's hard to say if you missed one.

I would also do `keras.config.set_random_seed(SEED)` just to make sure things are consistent across hosts.

examples/ml_perf/main.py
```diff
@@ -0,0 +1,242 @@
...
+                mode="fan_in",
+                distribution="uniform",
+                seed=SEED,
+            ),
+            optimizer=keras.optimizers.Adagrad(
+                learning_rate=embedding_learning_rate
+            ),
+            combiner="sum",
+            placement="sparsecore",
+            # TODO: These two args are not getting passed down to
```
hertschuh, line 68: You could try these XLA flags:
- `xla_sparse_core_max_ids_per_partition_per_sample`
- `xla_sparse_core_max_unique_ids_per_partition_per_sample`

examples/ml_perf/main.py
```diff
@@ -0,0 +1,242 @@
...
+            # `jax-tpu-embedding` properly, seems like.
+            max_ids_per_partition=max_ids_per_partition,
+            max_unique_ids_per_partition=max_unique_ids_per_partition,
+        )
+        feature_configs[f"{feature_name}_id"] = keras_rs.layers.FeatureConfig(
+            name=feature_name.replace("id", ""),
+            table=table_config,
+            # TODO: Verify whether it should be `(bsz, 1)` or
+            # `(bsz, multi_hot_size)`.
+            input_shape=(per_host_batch_size, multi_hot_size),
```
hertschuh, line 78: multi_hot as an input to embedding tables is not supported, it should be turned into a ragged or padded sequence of indices before being fed to the DistributedEmbedding.

abheesht17: > multi_hot as an input to embedding tables is not supported, it should be turned into a ragged or padded sequence of indices before being fed to the DistributedEmbedding.

Oh, I think our understanding of multi-hot is different. I am passing indices here. I think what you mean by multi-hot is `[[1, 0, 1, 0, 0], [0, 0, 1, 1, 1]]`

hertschuh (commented):
Some comments about the bash script

PR #146, opened 2025-08-22
Fix and clean up multihost preprocessing.

cantonios:
Multihost JAX programs require the same JAX program to run across all hosts.  To enforce this, we need to synchronize all input statistics, and repeat preprocessing if the stats have changed.

Also simplified the synchronization program for better efficiency by collecting all input statistics into a single JAX array.

### DIFF
--- a/keras_rs/src/layers/embedding/jax/distributed_embedding.py
+++ b/keras_rs/src/layers/embedding/jax/distributed_embedding.py
@@ -15,7 +15,6 @@ from jax_tpu_embedding.sparsecore.lib.nn import (
     table_stacking as jte_table_stacking,
 )
 from jax_tpu_embedding.sparsecore.utils import utils as jte_utils
-from keras.src import backend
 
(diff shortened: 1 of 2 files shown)


cantonios:
> > So the `jax.pmap(jax.lax.pmax(...))` was already in place before. So the core of the fix is the re-running of the preprocessing if the stats don't match?
> 
> Yes, that's the intention. I also merged the pmax calls into a single call rather than separate ones for each statistic, and removed a bunch of extra keras variables that it turns out we never actually needed.
> 
> > Were you able to test this. If so, how?
> 
> I'm working on a google-internal multiprocess test. It still fails though when the input size differs across hosts, so converting this PR to a draft.

Unmarking as draft.  This PR _should_ fix the issues with preprocessing, and simplifies the code a bit.  My local multiprocess test fails even without preprocessing, so there's some other deeper issue that needs to be addressed.

PR #156, opened 2025-09-24
Refactor: Use `embedding.EmbeddingVariables` in Keras-RS JAX embedding tests.

adityagupta1089:
This change updates the Keras-RS JAX embedding tests to use the `embedding.EmbeddingVariables` dataclass from `jax_tpu_embedding` for representing embedding tables and slot variables, instead of a custom tuple structure. This involves updating type hints, variable access, and build dependencies.

### DIFF
--- a/keras_rs/src/layers/embedding/jax/test_utils.py
+++ b/keras_rs/src/layers/embedding/jax/test_utils.py
@@ -7,6 +7,7 @@ import jax
 import keras
 import numpy as np
 from jax import numpy as jnp
+from jax_tpu_embedding.sparsecore.lib.nn import embedding
 from jax_tpu_embedding.sparsecore.lib.nn import embedding_spec
 from jax_tpu_embedding.sparsecore.lib.nn.embedding_spec import FeatureSpec
 from jax_tpu_embedding.sparsecore.lib.nn.embedding_spec import TableSpec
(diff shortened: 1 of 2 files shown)


keras_rs/src/layers/embedding/jax/test_utils.py
```diff
@@ -1,23 +1,25 @@
...
 import jax
-import keras
-import numpy as np
 from jax import numpy as jnp
+from jax_tpu_embedding.sparsecore.lib.nn import embedding
 from jax_tpu_embedding.sparsecore.lib.nn import embedding_spec
 from jax_tpu_embedding.sparsecore.lib.nn.embedding_spec import FeatureSpec
 from jax_tpu_embedding.sparsecore.lib.nn.embedding_spec import TableSpec
-
+import keras
```
cantonios, line 12: Can you move these imports back to where they were?

keras_rs/src/layers/embedding/jax/test_utils.py
```diff
@@ -311,14 +308,14 @@ def _create_samples(
 
 def stack_shard_and_put_tables(
     table_specs: Nested[TableSpec],
-    tables: Nested[jax.Array],
+    tables: T,
```
cantonios, line 311: Can we make this more specific?  I'm guessing there are tests or something that use jax.Array, but now others that use embedding.EmbeddingVariables?

cantonios (changes requested):

keras_rs/src/layers/embedding/jax/test_utils.py
adityagupta1089: Done.

hertschuh (approved):
Thanks for the PR!

PR #164, opened 2025-10-22
Fix CPU stats aggregation for JAX multihost.

cantonios:
Replaces custom pmap+pmax with the special-purpose multihost_utils.process_allgather.  Tested in a pseudo multihost (multiprocess) test.


abheesht17:
Thanks! This works for a dummy dataset on https://github.com/keras-team/keras-rs/pull/144. For the real dataset though, it freezes before training. Trying to work through this, maybe something wrong with my processing code

PR #167, opened 2025-10-29
Add heuristic for max ids of stacked tables.

hertschuh:
Propagate the `max_ids_per_partition` and `max_unique_ids_per_partition` from `TableSpec`s to `StackedTableSpec`s by taking the max from the stacked tables.


hertschuh:
@abheesht17 I verified this with the ml-perf example.

abheesht17 (approved):
LGTM, thanks!

PR #170, opened 2025-11-21
support all tests in layers to tpu

wenyi-guo:
support all tests in layers to tpu


keras_rs/src/layers/embedding/tensorflow/config_conversion_test.py
```diff
@@ -7,6 +8,10 @@
 from keras_rs.src.layers.embedding.tensorflow import config_conversion
 
 
+@pytest.mark.skipif(
+    keras.backend.backend() != "tensorflow",
+    reason="Backend specific test",
```
hertschuh: nitpick: can you say `reason="TensorFlow specific test",` to be more specific.

keras_rs/src/layers/feature_interaction/dot_interaction_test.py
```diff
@@ -8,10 +12,18 @@
...
 
 
 class DotInteractionTest(testing.TestCase, parameterized.TestCase):
     def setUp(self):
+        super().setUp()
+        if keras.backend.backend() == "tensorflow":
+            tf.debugging.disable_traceback_filtering()
+
+        self.on_tpu = "TPU_NAME" in os.environ
+        self._strategy = tpu_test_utils.get_tpu_strategy(self)
```
hertschuh: Let's not do that in each test class.

For `self.on_tpu`, you can probably do it once for all in `testing.TestCase.setUp`.

For `self.strategy`, let's make a  property that is lazily created, in `testing.TestCase.setUp`.

```python
@property
def strategy(self):
    if hasattr(self, "_strategy"):
        return self._strategy

    ...
    self._strategy = ...
    return self._strategy
```

Although, we may have to do that in a piece of code that is shared between all tests, like `conftest.py`.

keras_rs/src/layers/embedding/embed_reduce_test.py
```diff
@@ -9,8 +9,18 @@
...
+    import jax
+    from jax.experimental import sparse as jax_sparse
+except ImportError:
+    jax = None
+    jax_sparse = None
+
 
 class EmbedReduceTest(testing.TestCase, parameterized.TestCase):
+    def setUp(self):
+        super().setUp()
```
hertschuh: Revert, this is not needed.

keras_rs/src/metrics/dcg_test.py
```diff
@@ -19,6 +20,8 @@ def _compute_dcg(labels, ranks):
 
 class DCGTest(testing.TestCase, parameterized.TestCase):
     def setUp(self):
+        self._strategy = tpu_test_utils.get_tpu_strategy(self)
```
hertschuh: `super().setUp()` instead.

keras_rs/src/metrics/mean_average_precision_test.py
```diff
@@ -6,10 +6,13 @@
 
 from keras_rs.src import testing
 from keras_rs.src.metrics.mean_average_precision import MeanAveragePrecision
+from keras_rs.src.utils import tpu_test_utils
 
 
 class MeanAveragePrecisionTest(testing.TestCase, parameterized.TestCase):
     def setUp(self):
+        self._strategy = tpu_test_utils.get_tpu_strategy(self)
```
hertschuh: `super().setUp()` instead.

keras_rs/src/layers/retrieval/hard_negative_mining_test.py
```diff
@@ -9,6 +9,9 @@
 
 
 class HardNegativeMiningTest(testing.TestCase, parameterized.TestCase):
+    def setUp(self):
+        super().setUp()
```
hertschuh: You can revert this file.

keras_rs/src/utils/tpu_test_utils.py
```diff
@@ -0,0 +1,170 @@
...
+    ) -> Any:
+        del options
+        return dataset
+
+
+class JaxDummyStrategy(DummyStrategy):
+    @property
+    def num_replicas_in_sync(self) -> Any:
+        if jax is None:
+            return 0
```
hertschuh: I would remove this and simply `import jax` here. It should fail if jax is missing.

keras_rs/src/utils/tpu_test_utils.py
```diff
@@ -0,0 +1,170 @@
...
+                if jax is None:
+                    raise ImportError(
+                        "JAX backend requires jax to be installed for TPU."
+                    )
+                print("### num_replicas", jax.device_count("tpu"))
+                _shared_strategy = JaxDummyStrategy()
+            else:
+                _shared_strategy = DummyStrategy()
+            if _shared_strategy is None:
+                print("Failed to create the shared TPUStrategy.")
```
hertschuh: Remove, the code logic won't allow this

keras_rs/src/testing/test_case.py
```diff
@@ -1,12 +1,20 @@
...
+import tensorflow as tf
 
 from keras_rs.src import types
+from keras_rs.src.utils import tpu_test_utils
+
+StrategyType = Union[
+    tf.distribute.Strategy,
+    tpu_test_utils.DummyStrategy,
+    tpu_test_utils.JaxDummyStrategy,
+]
```
hertschuh: Use the one from `tpu_test_utils` instead of duplicating this declaration.

keras_rs/src/utils/tpu_test_utils.py
```diff
@@ -0,0 +1,127 @@
...
+
+class JaxDummyStrategy(DummyStrategy):
+    @property
+    def num_replicas_in_sync(self) -> Any:
+        import jax
+
+        return jax.device_count("tpu")
+
+
+StrategyType = Union[tf.distribute.Strategy, DummyStrategy, JaxDummyStrategy]
```
hertschuh: Because `JaxDummyStrategy` is a subclass of `DummyStrategy`, you don't actually need it here.

keras_rs/src/utils/tpu_test_utils.py
```diff
@@ -0,0 +1,127 @@
...
+) -> Any:
+    """
+    Final wrapper fix: Flattens allowed kwargs into positional args before
+    entering tf.function to guarantee a fixed graph signature.
+    """
+    if keras.backend.backend() == "tensorflow":
+        # Extract sample_weight and treat it as an explicit third positional
+        # argument. If not present, use a placeholder (None).
+        sample_weight_value = kwargs.get("sample_weight", None)
+        all_inputs = args + (sample_weight_value,)
```
hertschuh: Why is this needed? This was not in the original test.

And if it's needed, can this be made generic to handle any kwarg? You would need to take the values and then within `tf_function_wrapper` recreate the dict with the keys.

PR #172, opened 2025-12-12
Use `jax.shard_map` with JAX >= 0.8.0

hertschuh:
`jax.experimental.shard_map.shard_map` is going away.

### DIFF
--- a/keras_rs/src/layers/embedding/jax/embedding_lookup.py
+++ b/keras_rs/src/layers/embedding/jax/embedding_lookup.py
@@ -16,9 +16,30 @@ from jax_tpu_embedding.sparsecore.utils import utils as jte_utils
 from keras_rs.src.layers.embedding.jax import embedding_utils
 from keras_rs.src.types import Nested
 
-ShardedCooMatrix = embedding_utils.ShardedCooMatrix
-shard_map = jax.experimental.shard_map.shard_map  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


abheesht17 (approved):
Thanks!

keras_rs/src/layers/embedding/jax/embedding_lookup.py
```diff
@@ -16,9 +16,30 @@
 from keras_rs.src.layers.embedding.jax import embedding_utils
 from keras_rs.src.types import Nested
 
-ShardedCooMatrix = embedding_utils.ShardedCooMatrix
-shard_map = jax.experimental.shard_map.shard_map  # type: ignore[attr-defined]
+if jax.__version_info__ >= (0, 8, 0):
+    from jax import shard_map
+else:
+    from jax.experimental.shard_map import shard_map as exp_shard_map
```
abheesht17: Do you need `# type: ignore[assignment]` here too?

hertschuh: Not here because I'm importing it with a different name: `exp_shard_map`.

The `# type: ignore[assignment]` was needed in the other file because I'm importing them with the same name and they have a different signature. Here I had to write a wrapper to remap the `check_vma`, that triggers a different warning (`[misc]`).

PR #194, opened 2026-05-29
Fix for optimizer conversion test related to names.

hertschuh:
The change in Keras https://github.com/keras-team/keras/pull/22972 makes optimizer names unique. The name needs to be ignored in the optimizer equality test.


gaga1313:
Thanks for the fix!

gaga1313 (approved):

PR #202, opened 2026-09-03
Make all `actions/cache` actions in workflow consistent.

hertschuh:
This action was correct: https://github.com/keras-team/keras-rs/blob/main/.github/workflows/actions.yml#L160-L165

However, the other ones where inconstent in the files they used and the restore key.


laxmareddyp (approved):
Thanks.

PR #205, opened 2026-09-17
Fix tie handling in sort_by_scores

Nicholas022400701:
## What was wrong
`sort_by_scores` in `keras_rs/src/metrics/ranking_metrics_utils.py` hard codes `shuffle_ties=True` when it calls `get_shuffled_indices`. `RankingMetric.update_state` always passes a mask, so that branch is always taken and the `shuffle_ties` argument of the metric is never honored. `DCG`, `NDCG`, `MeanAveragePrecision`, `MeanReciprocalRank`, `PrecisionAtK` and `RecallAtK` therefore shuffle ties randomly even when constructed with `shuffle_ties=False`.
Example on the current `main` with any backend:
```python
metric = keras_rs.metrics.MeanReciprocalRank(shuffle_ties=False)
metric.update_state([[0.0, 0.0, 1.0, 0.0]], [[0.5, 0.5, 0.5, 0.5]])
metric.result()  # 1.0, 0.5, 0.333 or 0.25 depending on the run, expected 0.333
```
The torch workaround that subtracted `index * 1e-6` from the scores had two problems of its own:
1. `RankingMetric.update_state` pushes masked items only `keras.config.epsilon()` (1e-7) below the minimum score. The offset is ten times larger, so on torch a masked item at a lower index could outrank a valid item with the same score at a higher index. It also does nothing when the scores are large enough that subtracting 1e-7 is lost to float32 rounding, in which case masked items tie with valid items and the order given by the mask is the only thing separating them.
2. Scores that genuinely differ by less than `position * 1e-6` were reordered. With a list size of 1000 that is a gap of up to 1e-3.
## The fix
* Pass `shuffle_ties` through to `get_shuffled_indices`, which already handles the `False` case by sorting zeros. This matches `sort_by_scores` in `tensorflow_ranking`.
* On torch, use `ops.argsort(-scores)[:, :k]` instead of `ops.top_k`. The Keras torch backend implements `argsort` with `stable=True`, so tied scores keep the order given by `shuffled_indices` and the offset trick is no longer needed. JAX and TensorFlow keep using `top_k`, which is stable there.
* Document the `mask` argument and fix the comment that said masked values go to the beginning.
`ListMLELoss` also calls `sort_by_scores`, with `shuffle_ties=False` and `mask=None`, so it did not hit the shuffle branch and its results are unchanged.
## Tests
* New `ranking_metrics_utils_test.py` covering `sort_by_scores` with and without ties, masks, `k`, and `shuffle_ties` on both settings.
* Two new `MeanReciprocalRank` tests with `shuffle_ties=False`: tied scores keep their order, and a masked item that ties with valid items is ranked after them.
The full `pytest keras_rs/` suite and `pre-commit run --all-files --hook-stage manual` pass on the tensorflow, jax and torch backends. Since Actions are disabled on my fork and the upstream workflow uses self hosted runners, I ran the suite on GitHub hosted runners from a small helper repo that checks out this branch: https://github.com/Nicholas022400701/keras-rs-ci/actions/runs/35121627547


Nicholas022400701:
@googlebot I signed it!

PR #206, opened 2026-09-17
Fix RemoveAccidentalHits so accidental hits are actually removed

Nicholas022400701:
## What was wrong

`RemoveAccidentalHits` in `keras_rs/src/layers/retrieval/remove_accidental_hits.py` had two bugs.

1. The layer added `SMALLEST_FLOAT = ml_dtypes.finfo("float32").smallest_normal / 100.0` to the logits of accidental hits. That constant is about 1e-40. Adding it to any normal float32 logit is a no-op, so the layer returned the logits unchanged and accidental negatives kept their full softmax probability.

   ```python
   layer = keras_rs.layers.RemoveAccidentalHits()
   logits = [[1.0, 2.0, 3.0, 4.0]]
   labels = [[0.0, 0.0, 1.0, 0.0]]
   candidate_ids = [7, 5, 7, 8]  # candidate 0 has the same ID as the positive
   keras.ops.softmax(layer(logits, labels, candidate_ids))
   # main: [[0.032, 0.087, 0.237, 0.644]], the accidental hit keeps 3.2%
   # this PR: [[0.0, 0.090, 0.245, 0.665]]
   ```

2. The ID of the positive candidate was gathered with `ops.take(candidate_ids, positive_indices)` without an `axis`. `take` flattens its input when `axis` is `None`, so with per row `candidate_ids` of shape `[batch_size, num_candidates]` every row read its positive ID from row 0 and the wrong candidates were flagged.

The existing test did not catch either bug. It compared `out_logits` with `logits + SMALLEST_FLOAT`, which equals `logits` in float32, so both branches of the assertion passed no matter what the layer returned. It also indexed `candidate_ids` with the leading index of `logits` for rank 3 logits with rank 2 candidate IDs, while `candidate_ids` matches the trailing dimensions of `labels`.

## The fix

* Replace the constant with `MIN_FLOAT = ml_dtypes.finfo("float32").min / 100.0`, so accidental hits get a logit around -3.4e36 and a softmax probability of exactly zero. This is the same construction `HardNegativeMining` already uses for `MAX_FLOAT` in the mirror case, and the same value the TensorFlow Recommenders `RemoveAccidentalHits` layer uses.
* Gather the positive candidate ID with `ops.take_along_axis(candidate_ids, positive_indices, axis=-1)` so each row uses its own IDs. Non axis dimensions broadcast, so 1D `candidate_ids` and symbolic batch dimensions keep working.

## Tests

* `test_call` now checks against `MIN_FLOAT` and indexes `candidate_ids` on the trailing dimensions only, so the assertion is no longer vacuous.
* Two new tests, `test_call_1d_candidate_ids` and `test_call_2d_candidate_ids`, with hand computed expected logits and softmax probabilities for shared and per row candidate IDs.

The full `pytest keras_rs/` suite and `pre-commit run --all-files --hook-stage manual` pass on the tensorflow, jax and torch backends. Actions are disabled on my fork and the upstream workflow uses self hosted runners, so I ran the suite on GitHub hosted runners from a small helper repo that checks out this branch: https://github.com/Nicholas022400701/keras-rs-ci/actions/runs/35123016399


Nicholas022400701:
@googlebot I signed it!

